PR #20, opened 2026-07-25
fix(privacy): wire privacy control methods and correct the widen ceremony

TheGaySupreme123:
## Summary

Privacy was structurally unreachable, so the LLM provider path could never be exercised end to end. This wires it up and fixes four independent blockers found along the way.

| Fix | Why it was broken |
|---|---|
| Bind `privacy_get_effective` / `propose` / `tighten` handlers | `support_handlers` carried only observation handlers, so `ReadyApplication._support` raised `method_forbidden` for all three. |
| `vault.py` challenge purpose `privacy_policy_change` → `privacy_policy_widen` | Every issuing site (`human_control.py`, `unlock.py`, `elevated_bootstrap.py`) already used `_widen`; the map lookup returned `None`, so every privacy re-auth failed with `secret_purpose_mismatch`. |
| Data-category token rule in `confidential_protocol.py` | `PrivacyPolicyDecisionPreview.categories` was validated with `_require_token` (no underscores). Every `DataCategory` value has one, so the preview could never be constructed. |
| Implement `_LockedHumanEffects.decide_privacy` | Was an unconditional `raise HumanControlError("kind_forbidden")`. |

## Wire codec split

Catalog rows stay domain-shaped; control results must match the frozen `privacy-policy-1.0.0` schema (`local_sink_category_ceilings`, const `never_send`, decimal-string counters). Encoder and decoder are split, and the round trip is covered.

`provider_reconciliation` is written out by hand rather than reflected off the dataclass. The domain type names the generation `policy_generation` and pairs each digest with an internal reason string; the reviewed schema wants a decimal-string `policy_version` and digests only, under `additionalProperties: false`. A reflected body fails `validate_result` with `frame_invalid` — `test_domain_shaped_reconciliation_is_rejected_by_the_envelope` asserts that failure mode is real rather than theoretical.

## Scope on `privacy_get_effective`

`scope` is required by the frozen request schema. The CLI builds the machine scope from installation state and reports an unreadable id as a caller error instead of sending a body the service must reject.

## Tests

- `tests/unit/application/test_privacy_control.py` — every handler result body validated through the real `validate_result` envelope, plus the reflector regression guard.
- `tests/unit/adapters/privacy/test_privacy_policy_wire.py` — domain ↔ wire round trip, wire schema conformance, desired-state shape still decodes.
- `tests/builders/privacy_policies.py` — shared `local_only` / `minimal_external` policy builders.

## Verification

- `pytest tests/unit tests/property tests/conformance` — 1335 passed
- `ruff format --check` / `ruff check` — clean
- `pyright` — 0 errors
- `scan_public_boundary` / `generate_schemas --check` / `verify_resource_manifest --check` — PASS

## Not addressed here

- Policy-decision commitments are MAC'd under `yoetz/privacy-audit/local-approval/v1`, the local-*disclosure*-approval domain. It works and the store only persists the commitment, but it conflates two audit meanings; adding a dedicated domain touches the frozen installation domain set and deserves its own decision.
- `_send_secret` zeroes the caller's buffer, so a ceremony needing the passphrase twice (unlock, then privacy re-auth) would send zeros on the second use. Not reachable today because `prepare` requires `vault.ready`.
- No live provider smoke. This makes the path reachable; it does not prove an endpoint works.

## Summary by CodeRabbit

* **New Features**
  * Added privacy policy control operations for viewing effective policies, proposing changes, and tightening policies.
  * Added support for approving or denying privacy policy decisions non-interactively.
  * Added privacy policy export and wire-format compatibility.
  * Added machine-scoped privacy policy requests for CLI workflows and status reporting.
  * Added clearer handling of privacy reauthentication and decision outcomes.

* **Bug Fixes**
  * Improved validation of policy categories, scopes, and serialized policy data.
  * Corrected privacy policy transition loading and stale or expired decision handling.

* **Tests**
  * Added coverage for privacy policy serialization, control responses, and round-trip compatibility.


TheGaySupreme123:
Went through all five findings. Four were valid and are fixed in `e321dc9`, each with a test. One declined with reasoning.

**Fixed**

- **`daemon.py` — incomplete widening preview.** Agreed, and this was the most serious of the five: the preview is the human-consent artifact, and `scopes` was not a diff at all — it always reported the candidate's `effective_scope.kind` whether or not anything widened. A widen that only raised a local-sink ceiling produced an empty summary. Added `privacy_widening_summary` alongside `_is_tightening` so the summary a human approves and the comparison that decides a widen is required share one implementation, covering channel categories, local-sink ceilings, and per-channel `scope_ceiling`.

- **`catalog.py:404` — wire decode strictness.** Confirmed against the schema: `schema_version` and `never_send` are both in `required`, and `never_send` is a const deny list. Both are now mandatory. One deviation from your patch — `source["never_send"]` raises a bare `KeyError` on absence, while every other corruption path here raises the bounded `privacy_policy_row_corrupt`; my test caught that, so it uses `.get` with an explicit presence check.

- **`catalog.py:639` — reconstruct from `diff_canonical`.** Correct, and `diff_canonical` is `NOT NULL`, so it is always available. `prepare_transition` stores `canonical_encode({base_policy_digest, candidate_policy_digest})` and derives `exact_diff_digest` from it, so reloading reproduces the original digest exactly. Deriving it from the *current* policy meant the digest a human previewed was not necessarily the digest the commit was authorised against.

- **`privacy_control.py:182` — `expected_policy_version` binding.** Valid. Rather than dropping the second read, it is now verified against the generation and digest the proposal was prepared against; disagreement means the proposal is already stale, which the caller must resolve by re-reading.

**Declined**

- **`provider_status.py:135` — preserve installation-state errors.** Deliberate. This surface is a readiness *report* whose contract is to degrade per condition and name blockers, so an unreadable installation state reports `llm_inference_enabled: unknown` with a `state: "unknown"` blocker rather than a guess. That is pinned by `test_missing_installation_state_reports_unknown_not_a_false_ready`. Failing the whole report would also discard `semantic_enabled` and `endpoint_bound`, which are read from config and need no service call at all — strictly less information for an operator diagnosing exactly this situation.

Full bounded suite (unit, property, conformance, subprocess, integration): 1684 passed. ruff format/check clean, pyright 0 errors, `scan_public_boundary` PASS.

PR #39, opened 2026-07-27
fix: bind semantic provenance to the policy digest that authorized the dispatch

### LINKED ISSUE #38: Semantic provenance records all-zero policy digests on successful external dispatch (2026-07-27 run-2 dogfood)
## Summary

On the one genuinely successful external semantic check in the 2026-07-27 run-2 dogfood, the
recorded `SemanticProvenance` carried **all-zero** `policy_digest` and `privacy_policy_digest`,
while the agent-context privacy projection reported the real nonzero effective policy digest for
the same dispatch.

The authorization, request commitment, and privacy receipt still made the dispatch audit-linked, so
this is **not** a privacy failure. It is a provenance-quality failure: the recorded provenance
disagrees with the egress receipt about which policy authorized the physical dispatch.

## Root cause

Provider adapters mint placeholder policy digests, and every layer above passes them through
untouched:

| Site | `policy_digest` / `privacy_policy_digest` |
|---|---|
| `adapters/providers/openai_responses.py:402-403` | `"sha256:" + "0"*64` |
| `adapters/providers/openai_chat_completions.py:265-266` | `"sha256:" + "0"*64` |
| `adapters/providers/local_model.py:210-211` | `profile.release_resource_digest` (a *release artifact* digest, not a policy digest) |
| `adapters/providers/fake.py:81-82` | `_FAKE_DIGEST` (= all zeros) |
| `adapters/privacy/gateway.py:818-819, 840-841, 865-866` | `_ZERO_DIGEST` |

Run 2 used `fireworks-responses`, which routes to `openai_responses.py`.

Nothing catches it: `validate_sha256_digest` (`domain/values.py:456-459`) is a shape-only regex
(`^sha256:[0-9a-f]{64}$`), and all-zero is valid shape. `service/ready_composition.py:1324-1325`
passes the value straight into `SemanticProvenance`, and from there into the `check_recorded`
payload (`application/check.py:158-165`).

## The real digest is already in the adapter's hand

`SemanticEvaluatorPort.evaluate(case, deadline)` takes an `ApprovedProviderCase`, and both variants
already carry the real value — `ApprovedOutboundCase.policy_digest` (`domain/privacy.py:1243`) and
`ApprovedLocalDisclosureCase.policy_digest` (`domain/privacy.py:1289`). The gateway *proves* that
value equals the effective policy before dispatch, at `adapters/privacy/gateway.py:586-591`:

```python
if (
    case.policy_digest != authorization.policy_digest
    or registry.policy_digest != authorization.policy_digest
    or registry.service_generation != authorization.service_generation
):
    return PrivacyReason.AUTHORIZATION_STALE
```

The whole chain is equal by construction:

```
EffectivePrivacyPolicy.effective_digest -> ProviderRegistry.policy_digest
  -> EgressAuthorization.policy_digest -> ApprovedOutboundCase.policy_digest
  -> EgressReceipt.policy.policy_digest
```

## Proposed change

Make the gateway authoritative and stop adapters from minting policy digests at all.

1. **Gateway injection.** Immediately after `evaluator.evaluate(case, deadline)` returns (external
   dispatch at `gateway.py:544`, local dispatch at `gateway.py:725`), bind provenance
   `policy_digest`/`privacy_policy_digest` from `case.policy_digest`, following the existing
   `_with_request_commitment` (`gateway.py:781-784`) `replace(...)` pattern.
2. **Adapters stop asserting policy.** Thread `case.policy_digest` through the adapters'
   `_provenance` helpers so the value is never wrong even before the gateway overwrites it.
3. **A fence at the dispatch boundary.** Reject an all-zero policy digest on a *successful*
   dispatch. It belongs at the gateway/dispatch boundary, not in `ProviderAttemptProvenance`
   or `SemanticProvenance`, so legitimate all-zero construction in
   `tests/unit/domain/test_findings.py` keeps working.

Genuinely-blocked paths that never reached a provider (`_preconsume_result`,
`_unknown_outcome_result`, `_local_unavailable_result`, all `SemanticStatus.UNAVAILABLE`) are left
alone — zero there is arguably honest.

## Design gate

`CONTRIBUTING.md:28-37` lists **privacy and data-egress behavior** as design-gated, and this touches
`adapters/privacy/gateway.py`, so I am opening this before the PR rather than treating "it only
changes a digest value" as an exemption. Worth stating explicitly for the review:

- This **narrows** what a provider adapter is trusted to assert. It does not loosen any privacy
  boundary, widen any scope, or change what leaves the process.
- **No schema, manifest, or canonical-leaf change.** `_SEMANTIC_PROVENANCE_LEAVES` lists field
  *names* (unchanged), and `semantic-provenance-1.0.0.schema.json` types both fields as the
  shape-only `digest` `$ref`, so all-zero is schema-valid today and the fix is value-only. The
  fence is deliberately code-level to avoid a schema regeneration.

## Open question, deliberately not answered here

`policy_digest` and `privacy_policy_digest` are set to the **same value** at every site today, and
no second distinct digest is wired into this path. Binding both to `case.policy_digest` is honest
and matches the egress receipt. If a genuinely *distinct* privacy-policy digest is wanted, that is a
protocol decision needing an ADR, and it is out of scope here.

## Definition of done

1. On a successful external dispatch, provenance `policy_digest` and `privacy_policy_digest` equal
   `case.policy_digest`, which equals `receipt.policy.policy_digest`.
2. No all-zero digest can reach provenance on a successful dispatch, enforced by a test.
3. `tests/unit/domain/test_findings.py` passes unmodified — if it needed changing, the fence is in
   the wrong layer.
4. `SemanticStatus.UNAVAILABLE` blocked paths are unaffected.
5. No schema, manifest, or canonical-leaf regeneration in the diff.

Source of the observation: `docs/postmortems/2026-07-27-codex-testing-yoetz-openrouter-easy-linking-run2.md`.
One of three independent remediations from that run; siblings are #34 and #36.

TheGaySupreme123:
Closes #38. One of three independent remediations from the 2026-07-27 run-2 dogfood; siblings #35
and #37 are already merged. Shares no files with either.

## What was wrong

The one genuinely successful external semantic check in run 2 recorded provenance with **all-zero**
`policy_digest` and `privacy_policy_digest`, while the agent-context privacy projection reported the
real nonzero effective policy digest for the same dispatch.

The authorization, request commitment, and privacy receipt still made the dispatch audit-linked, so
this was **not** a privacy failure. It was a provenance-quality failure: provenance and the egress
receipt disagreed about which policy authorized the attempt.

Adapters minted placeholder digests and every layer above passed them through untouched. Nothing
caught it, because `validate_sha256_digest` is a shape-only regex and all-zero is valid shape.

The real value was **already in the adapter's hand**. `SemanticEvaluatorPort.evaluate(case, deadline)`
takes an `ApprovedProviderCase`, both variants carry `policy_digest`, and the gateway *proves* that
value equals the effective policy before dispatching (`gateway.py:586-591`, the staleness fence). It
was simply never plumbed into the `_provenance(profile, status, ...)` helpers.

## What changed

**The gateway is now the authority; adapters no longer assert policy at all.**

- **Gateway rebind.** `_with_policy_digest` binds provenance from `case.policy_digest` immediately
  after `evaluate()` returns, on both the external and local dispatch paths, following the existing
  `_with_request_commitment` `replace(...)` pattern. Provenance and
  `EgressReceipt.policy.policy_digest` now agree by construction, with the staleness fence as the
  proof.
- **Adapters stop lying at the source.** `case.policy_digest` is threaded through the `_provenance`
  helpers in `openai_responses`, `openai_chat_completions`, and `local_model`, so the value is never
  wrong even before the gateway overwrites it. `local_model` stops reporting
  `profile.release_resource_digest` — nonzero, but a *release artifact* digest is not a policy
  digest. `fake.py` rebinds its scripted placeholder at dispatch time, since a script is authored
  before any case exists.
- **A fence that cannot regress.** An all-zero digest on a `SUCCEEDED` result is refused. It lives at
  the **dispatch boundary**, not in `ProviderAttemptProvenance` or `SemanticProvenance` — so
  `tests/unit/domain/test_findings.py` and `fake.py`'s legitimate all-zero construction keep working
  unmodified. External dispatch degrades to the bounded `outcome_unknown` receipt, local to the
  bounded unavailable result; neither crashes and both stay receipted.

Genuinely-blocked paths that never reached a provider (`_preconsume_result`,
`_unknown_outcome_result`, `_local_unavailable_result`, all `SemanticStatus.UNAVAILABLE`) keep their
zeros. Nothing authorized them, so zero is the honest value there.

`normalize_response`, `classify_provider_failure`, and `normalize_local_response` gain a required
keyword-only `policy_digest`. These are adapter-internal public helpers with no callers outside the
adapters and their tests.

## Tests

- **Provenance carries the real digest** — both request-shape suites assert
  `result.provenance.policy_digest == case.policy_digest` and `!= "sha256:" + "0"*64`.
- **Provenance equals the egress receipt** — the invariant that matters, in
  `test_reconcile_and_dispatch_external_semantic_succeeds`, which already held both `result` and
  `receipt`: provenance == `receipt.policy.policy_digest` == `policy.policy_digest`.
- **End to end into the event** — `test_fake_provider_coordinator` asserts every scripted outcome
  reports the case's digest and that `_finalize` (which mirrors `ready_composition._provider_provenance`)
  carries it through to the recorded `SemanticProvenance`.
- **The fence itself** — pinned directly at the boundary that owns it. Reaching it through the public
  dispatch path is impossible by construction, since the gateway already requires case, authorization,
  and live registry to agree on the policy digest and a real one is a canonical digest. That is
  exactly why it is worth pinning: it is what stops the pre-fix behaviour from silently returning.
  The test also asserts an `UNAVAILABLE` result keeps its honest zeros.

## Design gate

`CONTRIBUTING.md:28-37` lists privacy and data-egress behavior as design-gated and this touches
`adapters/privacy/gateway.py`, so #38 was opened before this PR rather than treating "it only changes
a digest value" as an exemption. For the review:

- This **narrows** what a provider adapter is trusted to assert. It loosens no privacy boundary,
  widens no scope, and changes nothing about what leaves the process.
- **No schema, manifest, or canonical-leaf change.** `_SEMANTIC_PROVENANCE_LEAVES` lists field
  *names* (unchanged) and `semantic-provenance-1.0.0.schema.json` types both fields as the shape-only
  `digest` `$ref`, so all-zero was schema-valid before and this is a value-only fix. The fence is
  deliberately code-level to avoid a schema regeneration.

## Open question, deliberately not answered here

`policy_digest` and `privacy_policy_digest` are bound to the **same value**, as they were at every
site before this change; no second distinct digest is wired into this path. Binding both to
`case.policy_digest` is honest and matches the egress receipt. If a genuinely *distinct*
privacy-policy digest is wanted, that is a protocol decision needing an ADR, and it is out of scope
here.

## Verification

```
uv run pytest                    # 2052 passed, 17 skipped, 4 xfailed
uv run ruff check src tests      # All checks passed
uv run ruff format --check src tests
npx --no-install pyright         # 0 errors, 0 warnings, 0 informations
```

The full suite was run rather than just the touched areas, since the fence sits on the dispatch path.

### DIFF
--- a/docs/INTERFACES.md
+++ b/docs/INTERFACES.md
@@ -407,6 +407,12 @@ commitment when applicable; and the validated terminal status/reason pair. The e
 and wire conversions are frozen in `domain/findings.md` and
 `semantic-provenance-1.0.0.schema.json`.
 
+`policy_digest` and `privacy_policy_digest` on `ProviderAttemptProvenance` and
+`SemanticProvenance` are bound by the outbound gateway to the effective policy digest that
+authorized the physical dispatch — the same value carried by `ApprovedOutboundCase.policy_digest`,
+`EgressAuthorization.policy_digest`, and `EgressReceipt.policy.policy_digest`. A provider adapter
+never asserts them. On a successful dispatch they are never placeholder or all-zero values.
+
 A deterministic candidate/finding forbids provenance. A semantic-model-derived candidate/finding
 requires receipt-finalized provenance whose status/reason is exactly
 `succeeded/semantic_completed`; failed, refused, stale, late, invalid, timeout, or unavailable
--- a/src/yoetz/adapters/privacy/gateway.py
+++ b/src/yoetz/adapters/privacy/gateway.py
@@ -46,6 +46,7 @@ from yoetz.domain.privacy import (
     PRIVACY_REQUEST_COMMITMENT_ALGORITHM,
     ApprovedLocalDisclosureCase,
     ApprovedOutboundCase,
+    ApprovedProviderCase,
     ConsentSource,
     EgressAuthorization,
     EgressChannel,
@@ -542,6 +543,7 @@ class PolicyEnforcingOutboundGateway(OutboundGatewayPort):
         dispatch_started_at = self._clock.now_utc()
         try:
             result = await evaluator.evaluate(case, deadline)
+            result = _with_policy_digest(result, case)
             result = _with_request_commitment(result, commitment)
             outcome, receipt_reason = _result_outcome(result)
         except Exception:  # noqa: BLE001 - an ambiguous transport failure never leaks native text
@@ -722,7 +724,12 @@ class PolicyEnforcingOutboundGateway(OutboundGatewayPort):
         except Exception:  # noqa: BLE001 - a lost consume race is bounded here
             return _local_unavailable_result(case)
 
-        return await evaluator.evaluate(case, deadline)
+        result = await evaluator.evaluate(case, deadline)
+        # Only the rebind is fenced here: an evaluator failure keeps propagating exactly as before.
+        try:
+            return _with_policy_digest(result, case)
+        except ValueError:
+            return _local_unavailable_result(case)
 
     # -- lifecycle ------------------------------------------------------------------------------
 
@@ -782,6 +789,36 @@ def _with_request_commitment(result: SemanticResult, commitment: str) -> Semanti
     """Carry the gateway-bound request commitment into the receipt-bound semantic outcome."""
 
     return replace(result, provenance=replace(result.provenance, request_commitment=commitment))
+
+
+def _with_policy_digest(result: SemanticResult, case: ApprovedProviderCase) -> SemanticResult:
+    """Bind provenance to the policy digest that actually authorized this physical dispatch.
+
+    The gateway, not the provider adapter, is the authority on which policy admitted an attempt: it
+    already refuses to dispatch unless ``case.policy_digest`` equals both the authorization's and
+    the current registry's policy digest (see :meth:`OutboundPrivacyGateway._predispatch_reason`),
+    and it stamps that same value onto :class:`EgressReceipt`. Rebinding here makes provenance and
+    the egress receipt agree by construction, and keeps a foreign adapter from ever asserting what
+    authorized it.
+
+    A dispatch that reached a provider and succeeded must carry a real digest. An all-zero
+    placeholder surviving the rebind means the authorizing policy digest was itself a placeholder,
+    so the attempt is refused rather than recorded as if a real policy had authorized it: external
+    dispatch degrades to the bounded ``outcome_unknown`` receipt and local dispatch to the bounded
+    unavailable result. Genuinely-blocked results that never reached a provider keep their
+    :attr:`SemanticStatus.UNAVAILABLE` zeros untouched.
+    """
+
+    provenance = replace(
+        result.provenance,
+        policy_digest=case.policy_digest,
+        privacy_policy_digest=case.policy_digest,
+    )
+    if provenance.status is SemanticStatus.SUCCEEDED and (
+        provenance.policy_digest == _ZERO_DIGEST or provenance.privacy_policy_digest == _ZERO_DIGEST
+    ):
+        raise ValueError("privacy_gateway_provenance_policy_digest_placeholder")
+    return replace(result, provenance=provenance)
 
 
 def _result_outcome(result: SemanticResult) -> tuple[PrivacyOutcome, PrivacyReason | None]:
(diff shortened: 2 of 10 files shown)


TheGaySupreme123:
@codex please review pr

Issue #141, opened 2026-08-05
[test]: Complete the awaiting_human state-machine failure matrix


TheGaySupreme123:
## Duplicate search

- [x] I searched existing issues and PRs and did not find a duplicate.

Related, not duplicates:

- #132 owns the dogfood **preflight and provenance gate** (experiment protocol). It does not own the
  runtime ceremony defect below, so it cannot close on its own.
- #139 / PR #140 landed the standing bounded **Assisted** default. That work made the standing-policy
  path usable; it did not make the `confirm_every_request` path resumable.
- PR #103 repaired informed widening previews. PR #39 binds semantic provenance to the authorizing
  policy digest. #69 covers frozen privacy-selected semantic case content.

No existing issue owns "a check that reaches `awaiting_human` must stay resumable, and the local
disclosure ceremony must work in the vault mode users actually run."

## Problem statement

Under `confirm_every_request`, a semantic `check` that needs a local disclosure decision is currently
unrecoverable. Four independent defects compose into one dead end. All four are verified against
`f7ea710cd26acba446e52c0bb0254816a8138589` (current `main`).

### 1. The local disclosure ceremony is forbidden in keyring mode

`src/yoetz/service/daemon.py:1647` and `src/yoetz/service/daemon.py:1800` both gate the
`PrivacyPendingTarget` path on passphrase mode:

```python
if not self._vault.ready or mode != "passphrase":
    raise HumanControlError("kind_forbidden")
```

The disclosure branch added for #139 sits *after* that guard, so `yoetz privacy decide-disclosure
<pending_id>` returns `kind_forbidden` on any keyring-mode installation — which is the normal
production configuration. Recording an already-bounded disclosure decision does not widen policy and
should not require passphrase mode or strong reauthentication. Policy widening, credential changes,
and other elevated operations must keep their existing stronger authority requirements.

### 2. The pending ID never reaches the caller

`SemanticEgressAwaitingHuman` carries `privacy_proposal_id` and `expires_at`, but
`_map_egress_to_final` (`src/yoetz/service/ready_composition.py:1551`) drops both:

```python
if type(result) is SemanticEgressAwaitingHuman:
    return FinalSemanticEvaluation(
        SemanticStatus.AWAITING_HUMAN, SemanticReason.HUMAN_APPROVAL_REQUIRED
    )
```

The agent is told a human must approve, but is not told *what* to approve or by when. There is no
[...]

Issue #132, opened 2026-08-04
[change]: Add a policy-enabled semantic dogfood preflight and provenance gate


TheGaySupreme123:
## Summary

The retained dogfood task `019fc915-a0bd-7803-b5d9-d8cbb9c65981` was used to assess how well Yoetz gives qualitative feedback, but its MCP route explicitly declared:

```text
Route profile: strict. This route will not request external semantic review for this process lifetime.
```

The final check returned:

```text
semantic_status: blocked_by_policy
semantic_reason: route_semantic_ceiling
semantic_provenance: null
```

This was correct privacy behavior. The strict route returned before provider capability evaluation or dispatch, and the receipt honestly stated that no provider attempt or semantic finding was recorded.

The experiment problem is that a strict route cannot measure semantic usefulness. We need an explicit dogfood preflight and gate that distinguishes:

- **privacy correctness:** strict route performs zero semantic egress;
- **semantic availability:** an explicitly selected policy route is configured and authorized;
- **semantic execution:** an actual attempt has receipt-bound provenance;
- **semantic usefulness:** feedback changes or challenges the agent's work.

This issue is an experiment/setup change. It must not silently widen default policy or turn dogfood into an egress bypass.

## Duplicate search

Searched issues and PRs for strict route, `route_semantic_ceiling`, semantic-mode selection, policy-route dogfood, provider provenance, and semantic preflight.

Related, not duplicates:

- PR #18 repaired semantic-mode selection/disclosure and made deterministic-only limitations visible.
- #13 / PR #14 cover privacy-gated semantic advice and observation routing.
- #15 and provider-related work cover provider dispatch/capability.
- PR #39 binds semantic provenance to the authorizing policy digest.
- #69 covers frozen privacy-selected semantic case content.
- #128/#129 cover schema/feedback authorability, not egress.

No existing issue owns a controlled dogfood preflight that refuses to score semantic quality unless the route, authorization, provider readiness, and actual attempt provenance are all demonstrated.

## Evidence and current behavior

Observed commit: `959f20c36b0d50b21a482b85035adbe5fc8848d4`.

The implementation matches the session result:

- `yoetz mcp serve --semantic off` maps to a strict bridge runtime.
- the strict harness command is `yoetz mcp serve --semantic off`;
- setup chooses strict when semantic verification is disabled or no provider/local model is configured;
- noninteractive/deferred and local-only review registration choose strict;
[...]

PR #142, opened 2026-08-05
Make semantic checks authorization-aware and resumable across disclosure decisions

TheGaySupreme123:
## Issue

- Linked issue: `Refs #141`, `Refs #132`
- [x] I searched existing issues and pull requests for duplicates before opening this PR.
- [x] This change **is** design-gated (protocol, privacy/egress, storage/durability, packaging, ADR).
  Maintainer acknowledgement is recorded on
  [#141](https://github.com/TheGaySupreme123/yoetz/issues/141#issuecomment-5191983259) before any
  code was written.

## Summary

A semantic `check` that needed a local disclosure decision was unrecoverable. Five defects composed
into one dead end, all verified against `f7ea710` before any change:

1. `yoetz privacy decide-disclosure` returned `kind_forbidden` on every **keyring-mode**
   installation, because the ceremony was gated on passphrase mode. Keyring is the ordinary
   production configuration, so the `confirm_every_request` posture had no way to be answered at all.
2. The pending proposal id and its expiry were **discarded** by `_map_egress_to_final`. The caller
   was told a human must approve, but not what to approve or by when.
3. `awaiting_human` is not a retriable status, so it fell through to the exhaustion path: the
   physical attempt was written `FAILED`, the semantic job terminalized, and the check committed a
   terminal result. Even a granted approval then had nothing left to resume.
4. The CLI reported all of this as `service_unavailable`, sending the operator to restart a service
   that was running correctly.
5. The agent-facing text said nothing about any of it, so on hitting a pending decision the agent
   read Yoetz's SQLite catalog and product source looking for the id, issued a fresh check request,
   and declared completion before semantic closure.

This PR makes `awaiting_human` a real, resumable, nonterminal state end to end.

### Durable state machine

`awaiting_human` keeps the check operation, the semantic job, and the physical attempt nonterminal.
No provider was reached and the case was never sent, so the attempt stays `started` and the job stays
`leased`, with the suspension recorded in a new `semantic_disclosure_waits` table.

`claim_semantic_job` consults a live wait **before** the lease-reclaim path. This is the subtle part:
the provider request id is part of the prepared bytes the human approves. A human can easily take
longer than the 60s lease TTL, and the reclaim path would expire the started attempt and mint a new
provider request id — changing the prepared case digest and silently orphaning the decision. The
bound on the wait is the proposal's own expiry, not the lease.

The wait is one-use, so denial or expiry resumes exactly once. A transient provider retry creates a
fresh proposal needing a fresh decision. Resuming an approved proposal records
`per_request_local_human`, never a rewrite to baseline-policy consent.

### Confidential ceremony

Disclosure decisions are allowed whenever the vault is ready, including keyring mode. Answering a
`confirm_every_request` disclosure decides one exact prepared case already bounded by the committed
policy; it can neither widen the policy nor authorize any other case. **Policy widening, credential
set/rotate, and every other elevated operation keep their existing stronger authority requirement**,
with a test asserting a policy decision is still refused in keyring mode.

Before recording, the ceremony validates proposal existence, expiry, policy generation/digest,
destination commitment, prepared case digest, and one-use state. Structural failures map to
`pending_unavailable`, `pending_not_actionable`, and `ceremony_unsupported` instead of one
undifferentiated `target_invalid`. Absent and expired deliberately share one reason so an unknown id
is never confirmed.

### Public interface (pre-release 0.1 contract correction)

`check-result-1.0.0` gains a nonterminal success branch discriminated on `state`:

- `state: "awaiting_human"`, `semantic_status: "awaiting_human"`,
  `semantic_reason: "human_approval_required"`
- a typed `continuation` with `kind`, `pending_id`, `expires_at`, the fixed
  `["yoetz","privacy","decide-disclosure",pending_id]` command, `replay_request_id`, and instruction
- **no verdict, findings, coverage, or semantic provenance** — a completion-grade shape would let a
  caller conclude from a check that never ran

`execute_check_commit` returns this before ranking, phase advance, or commit. Committing is exactly
what made an approved decision useless.

`status(view="operation")` returns the same continuation for a pending check and `null` for every
other state, read from the durable wait rather than from anything the caller kept. That avoids a
seventh MCP tool while supporting recovery after lost output or compaction.

Version numbers are unchanged as a pre-release 0.1 correction; conformance fixtures and
`docs/INTERFACES.md` move with the contract.

### Agent-facing text

The skill, server initialize instructions, and `check` descriptor now state that an active semantic
route is a bounded standing policy the owner selected during setup and that `check` **cannot widen**
it; that host authorization and a Yoetz disclosure decision are different permissions; that
`awaiting_human` is nonterminal and carries a continuation to follow rather than a coverage gap to
disclose; and that a published completion claim is an assertion awaiting verification.

`openWorldHint` stays `true` on the policy route, with a test asserting it. Weakening a truthful
annotation to win an auto-review classifier would be a lie told to a safety check.

## Documentation

- Behavior change? **yes**
- Docs updated:
  - `docs/adr/ADR-009-data-egress-privacy.md` — answering a `confirm_every_request` disclosure is
    explicitly *not* a loosening operation and needs only a ready vault; the human-involvement table
    says "on any ready vault".
  - `docs/INTERFACES.md` — `awaiting_human` documented as the one nonterminal semantic status, with
    the one-use wait semantics.
  - `guidance/agent-instructions.md`, `skills/codex/yoetz/SKILL.md` — new sections, regenerated into
    packaged copies and the committed `.agents/plugins/yoetz/` tree.

## Verification

Commands run:

```text
uv run pytest tests/unit tests/integration -q      # 2189 passed
uv run pytest tests/packaging -q                   # 235 passed, 3 skipped, 4 xfailed
uv run pytest tests/conformance -q                 # 219 passed
uv run pytest tests/tui -q                         # passed
uv run ruff check .                                # All checks passed
uv run ruff format --check .                       # clean
npx --no-install pyright                           # 0 errors, 0 warnings
uv run python scripts/generate_schemas.py --check  # PASS (58 schemas match)
uv run python scripts/verify_resource_manifest.py --sync   # PASS (83 resources)
```

Known unrelated failure: `tests/property/test_service_control_frames.py::test_valid_hello_frames_round_trip_canonically`
fails on this machine with a Hypothesis `DeadlineExceeded` (690ms first run, 0.60ms on rerun).
Confirmed failing at `origin/main` in a clean worktree, so it predates this branch.

## Boundaries

- [x] No material from `docs/architecture/` or other ignored private drafting inputs is included.
- [x] I will disposition every human and code-review-agent comment before merge: fix it, or reply
      explaining why it is invalid or out of scope.

## Notes for reviewers

**Migration 0005 is additive on purpose.** The `semantic_jobs` state CHECK lives on a STRICT table
with circular foreign keys to `semantic_attempts`, and `run_migrations` executes every migration
inside one transaction with `PRAGMA foreign_keys = ON` — which SQLite will not let you toggle there.
A 12-step table rebuild has no safe place in that upgrade path, so the wait is a side table and the
suspended job stays `leased`.

**`check-result` and `status-result` are hand-authored schemas, not generated**, despite having
registry loaders. `generate_schemas.py --check` validates committed bytes rather than diffing against
the model, so a hand-maintained file passes `--check` — and `--write` silently replaces it with a
differently shaped pydantic-derived document. Both were edited by hand here, then their
`schemas/manifest.json` member digests recomputed, `sync_service_status_schema.py` run, and
`verify_resource_manifest.py --sync` run.

**Not in scope, still open on #141:** the remaining state-machine and production-composition test
matrix, packaging/install verification, and the `codex-testing` dogfood campaign. The dogfood is the
only way to answer whether `auto_review` accepts the truthful descriptor; per the acceptance
criteria this PR should not be treated as complete until that trial runs.


TheGaySupreme123:
## Review disposition — `1f1d8e3`

Every CodeRabbit finding verified against the code before acting. Nine fixed, two declined, one
deferred as a separate pre-existing issue.

### Fixed — real defects

| # | Finding | What was actually wrong |
| --- | --- | --- |
| [12] | `ready_composition.py` — continuation used the wrong request id | **The most serious one.** On the durable path `_map_egress_to_final` built the continuation from `result.request_id`, which is `handle.provider_request_id`. `CheckAwaitingHuman.__post_init__` requires `continuation.request_id == self.request_id`, so production raised a ledger validation error instead of returning the suspension. My tests missed it because they only exercised the protocol models and the runtime-is-`None` path, where the candidate id happens to be the operation id. Now the check operation id is threaded through explicitly. |
| [8] | `service.py` — projection merges `privacy_projection` | `project_result_for_client` adds `privacy_projection` to every CHECK success body before `_public_model`. `CheckAwaitingHumanModel` did not declare it, `_ClosedModel` forbids extras, and the schema branch sets `additionalProperties: false`. A suspended check reaching the projection path failed as a projection error. Fixed by declaring `privacy_projection` on the branch and in the schema rather than special-casing the projection: every projected success body carries one, and the nonterminal branch should not be the exception. |
| [11] | `daemon.py` — `_human_error_code` allowlist | Confirmed by reading: the allowlist had none of `ceremony_unsupported`, `pending_unavailable`, `pending_not_actionable`, so all three became `internal_error` at the daemon boundary and the CLI reporting work in `99e7c83` could never see them. |
| [4] | `repository.py` — upsert `COALESCE(resolved_at)` | A fresh attempt superseding a *resolved* wait would write `state='awaiting'` beside a non-null `resolved_at`, which the 0005 table CHECK rejects. All identity and pending fields now come from `excluded`; only `created_at` survives. |
| [5] | `repository.py` — waits not restored on restart | Correct, and it is the same `_ensure_recovered` gap already recorded on #141. `semantic_disclosure_waits` rows are now hydrated before `_requires_recovery` clears, so a restart no longer answers "no continuation" for a check that is genuinely suspended. |
| [6] | Duplicated instruction prose | Parity-critical and correct: the check result and the recovery query must return the *same* continuation. Now one `DISCLOSURE_CONTINUATION_INSTRUCTION` in `domain/values.py` beside `DISCLOSURE_CONTINUATION_KIND`, imported at both sites. |
| [13] | Byte-parity loop stopped at bundle 0003 | Correct, and pre-existing — 0004 was already unchecked before this branch. The loop now enumerates `BUNDLE_MIGRATIONS` / `CATALOG_MIGRATIONS` so it cannot fall behind again. |
| [2] | Guidance overclaims continuation recovery | Fair. `status.py` treats recovery as best effort, so the guidance promising `status` "returns the same continuation" was stronger than the code. Both copies now say it returns the continuation *while the durable record is available*, and to re-run the check rather than guess an id when it is not. |
| [1] | Checklist said "replay the same `request_id`" | Correct that this reads as preserving only the identifier. Now "replay the exact same `check` request with the same `request_id`", matching the body text. |

Packaged skill, resource manifest, `support/runtime-support.json` digests, and the committed
`.agents/plugins/yoetz/` tree were regenerated through their owning paths (a real `install_plugin`
run, not a hand edit).

### Declined

**[9] `except TypeError, ValueError:` is invalid Python 3 — marked 🔴 Critical.**
Not correct here. `pyproject.toml` pins `requires-python = ">=3.14,<3.15"`, and
[PEP 758](https://peps.python.org/pep-0758/) allows unparenthesized `except` tuples from Python 3.14.
The form is already used in roughly thirty places across `src/yoetz/` (for example
`cli/app.py:313`), the module imports fine, and the whole suite passes. Changing it would make this
file inconsistent with the rest of the codebase.

**[10] `schemas/manifest.json` is 6,628 bytes smaller than `src/yoetz/resources/manifest.json`, so one is stale.**
These are two different documents, not a mirror pair. `schemas/manifest.json` has keys
`{manifest_schema, manifest_version, members}` and inventories schema documents;
`src/yoetz/resources/manifest.json` has `{entries, package, resource_set_digest, resource_set_version, schema}`
and inventories all 83 packaged resources. A size difference between them carries no information.
`verify_resource_manifest.py --sync` passes and the byte-parity tests confirm the actual mirror
(`schemas/` → `src/yoetz/resources/schemas/`) is identical.

### Deferred — real, but pre-existing and out of scope

**[0] `.yoetz-plugin-install.json` and `skills/yoetz/manifest.json` disagree about `SKILL.md`.**
Verified: the install marker records the rendered `SKILL.md` at 13,560 bytes / `f6aabb…`, which
matches the canonical `skills/codex/yoetz/SKILL.md` exactly. The compatibility manifest's
`managed_members` records 9,502 bytes / `08ef013…`. But `git show origin/main` shows the *same*
9,502 / `08ef013…` values there, against a differently sized SKILL.md — so the compatibility
manifest's member digests were already stale before this branch and are not something this PR
introduced or widened. Two manifests describing the same logical file with different digests is a
genuine problem worth fixing, but it needs its own issue and an owning regeneration script rather
than a hand edit inside this PR.

### Not reproduced

**[3] `continuation` required in the status schema vs `exclude_unset` dumps.** The concern is real in
principle — `StatusOperationPageModel.model_dump(exclude_unset=True)` does drop `continuation`. But
the actual status path builds the page through `_dump_closed_omitting_optional_nulls`, which keeps
required nullable leaves as explicit null, and every conformance status test passes. Left as
required-and-nullable to match the established pattern for `next_cursor` and `semantic_provenance`.

**[7] Handle disclosure-wait persistence failures before returning.** A `record_disclosure_wait`
failure escaping between dispatch and return would leave the attempt `started` and the job `leased`.
That is now the *intended* resting state for a suspension, and the operation lease still expires, so
the ordinary reclaim path recovers it. Hardening this against injected storage faults is worth doing
with the rest of the state-machine test matrix that #141 still tracks, not as an untested change here.

### Verification after these fixes

```text
uv run pytest tests/unit tests/integration -q   # 2193 passed
uv run pytest tests/conformance -q              # 219 passed
uv run pytest tests/packaging -q                # 235 passed, 3 skipped, 4 xfailed
uv run ruff check . && ruff format --check .    # clean
npx --no-install pyright                        # 0 errors
```

Issue #206, opened 2026-08-11
[change]: Check cadence guidance — nudge a check at subtask transitions, not only at the completion claim


TheGaySupreme123:
## Duplicate search

- [x] Searched issues/PRs for check cadence, mid-run, guidance timing. No duplicate. Related context: #187 (check pause semantics), the cadence table itself.

## Area

Guidance / skill wording (`guidance/workflow.md`, `skills/codex/yoetz/SKILL.md`). Enhancement.

## Problem: the guidance back-loads `check`, so feedback arrives when it is most expensive to act on

Current cadence table (`guidance/workflow.md`, anchor `#cadence`):

> `check` | After publishing the completion claim and its evidence, and again after any material edit or new evidence. …

In the 2026-08-11 dogfood (session `019ff241`), the agent followed this **exactly**: plan published at 19:22Z, ~20 minutes of material work (content rewrite, an unplanned Astro 5→7 major upgrade, browser verification), four catch-up `publish_work` batches, then the session's **single** `check` at 19:46Z — immediately before claiming done. The check surfaced a real, actionable gap (`evidence_content_digest_only`) — but at claim time the only economical exit was "accept the gap in the receipt". Had a check run at the first phase boundary (~19:25Z), the same finding would have arrived while course-correcting was cheap (capture evidence differently, or scope the claim), and the scope expansion (Astro 7) would have been checked against a plan it wasn't in.

The agent was compliant; the cadence is the defect. `publish_work` is prescribed per material transition, but `check` is prescribed only at the end — so the feedback loop's one high-value probe fires exactly once, at the point of maximum sunk cost.

## Proposal (maintainer-selected: nudge, mode stays with the agent)

Add to the `check` cadence row (workflow.md + SKILL.md + the checklist), in substance:

> Also consider a check when you move between subtasks or phases — after publishing that transition's batch — not only at the completion claim. Choose the mode deliberately: `deterministic_only` is local and fast and catches record-hygiene gaps (stale ledger, digest-only evidence, open obligations) early; reserve semantic review for the claim unless the transition itself warrants it.

Explicitly a nudge, not a mandate — "A check with no new events since the last one adds nothing" stays as the guard against ritual checking. The cost asymmetry is worth stating in the guidance so agents can weigh it: this session's semantic check cost a 27.5 s provider round trip and a 60 KB disclosure; a `deterministic_only` check is milliseconds and never leaves the machine.

## Also in scope: cadence-table drift between the two sources

[...]

Issue #205, opened 2026-08-11
[change]: Default evidence intake — let diligent runs clear evidence_content_digest_only (transcript import, captured content, legibility)


TheGaySupreme123:
## Duplicate search

- [x] Searched issues/PRs for evidence intake, captured content, transcript import, coverage ceiling. Prior art: #176 (closed COMPLETED via #178 — its *minimal* fix landed; its explicitly-deferred "larger option" is route 3 below), #177 (4096-byte case cliff). The observation-delivery bug #204 is a prerequisite for route 2.

## Area

Evidence intake / coverage. Enhancement, design-gated in parts.

## Problem: a maximally diligent agent cannot clear the evidence gap

2026-08-11 dogfood (session `019ff241`, yoetz.dev landing deploy): the agent held genuinely strong evidence bytes — `curl -sS -I https://yoetz.dev/` returning `HTTP/2 200` with HSTS, `vercel inspect` showing `● Ready`, browser evals returning `{"noHorizontalOverflow":true,"errors":[]}` — all present verbatim in the Codex rollout. **None of those bytes had any channel into yoetz.** The only option was to paraphrase in prose, `shasum` its own sentence, and publish `strength: content_digest` with `provenance: caller_asserted` — a digest of a summary, not of evidence.

The check (correctly) fired `evidence_content_digest_only` on exactly this, and "accept the gap" was the only available exit: over cooperative MCP, `content_availability: captured` is unreachable by construction — only the observation coordinator ever creates `CAPTURED_CONTENT` objects (`observation_coordinator.py:566, 803, 1222`), the six MCP tools include no upload, and digest-only evidence adds a gap code without moving any coverage dimension (`deterministic_checks.py:1353-1369, 1505-1536`). Net effect: **a perfect run and a lazy run both terminate at `unresolved_findings_remain`.** The honesty is correct; the ceiling is the problem.

## Scoped routes (each independently shippable; maintainer-selected)

[...]

PR #213, opened 2026-08-12
Observation delivers and stops taxing the session: hook latency loop (#209, #210, #211)

### LINKED ISSUE #209: [bug]: Every observation hook exceeds its own 3s timeout — sync PreToolUse/PostToolUse add ~6s per tool call and are killed
## Duplicate search

- [x] Searched open issues/PRs for hook, observe, drain, outbox, latency, timeout, perf. No duplicate. #204 (closed by #208) fixed the *"hooks never fire"* leg; this is a new, post-#208 regression in the *"hooks fire but never finish"* direction. #203 is resource transport, not hook execution.

## Area

Observation subsystem: Codex lifecycle hook execution budget. `adapters/integrations/codex_plugin.py` (hook declaration) + `cli/observe_hooks.py` (handler).

## Observed — measured 2026-08-12 on this machine, current `main` (`dff08d5`)

`adapters/integrations/codex_plugin.py:104` declares `observe_timeout = 3` for every `yoetz hooks observe` handler. The handler cannot meet it. Wall-clock, real stdin payloads, installed `yoetz` on `PATH`:

| Event | Declared timeout | Measured wall time |
| --- | --- | --- |
| `SessionStart` | 3s | **4.89s** |
| `PreToolUse` | 3s | **3.06s** |
| `PostToolUse` | 3s | **3.20s** |
| `Stop` | 3s | **3.19s** |

Every observation hook overruns. `PreToolUse` and `PostToolUse` are declared **synchronous** (Codex defaults `async: false`), so **every tool call in a Codex session pays ~6s of added latency and then has both hooks killed at the deadline** — the user-visible symptom that stopped dogfood session `019ff71d-4132-7563-98b6-20d403dc94ed` after 75 seconds.

The cost is not interpreter startup: bare `yoetz --version` is **0.29s** and `import yoetz.cli.app` is **0.26s**. It is work inside the handler. `cProfile` on one `PreToolUse` invocation:

```
11.053s total (profiler-inflated; ~3.1s unprofiled)
  10.857  cli/observe_hooks.py:680(handle_observe)
   6.714  protocol/canonical.py:273(_canonical_text)      585,920 calls
   5.367  protocol/canonical.py:336(_encode_string)       692,012 calls
   4.907  adapters/integrations/observation_local.py:1374(_load)   << 17 calls
   3.872  cli/observe_hooks.py:502(_drain_outbox)
```

Two multipliers:

1. **`_load` runs 17× per hook invocation**, each re-reading and re-parsing the whole workspace state file. That file is currently **383,926 bytes**; `_save` canonically re-encodes it (**36ms/encode**) on every outbox row mutation. Cost scales with store size, and the store only grows (see the quarantine-growth issue filed alongside this one).
2. **`_drain_outbox` adds `_HOOK_DRAIN_BUDGET_SECONDS = 1.75`** (`observe_hooks.py:95`) *on top of* that store I/O. The 1.75s budget bounds only the drain loop, not the load/save around it, so 1.75 + ~1.4s of store work already exceeds 3s before any useful work happens.

## Codex-side facts (openai/codex, `codex-rs/hooks`, `codex-rs/config/src/hook_config.rs`)

Checked against the host implementation, because the fix depends on it:

- `#[serde(default, rename = "timeout")] timeout_sec: Option<u64>` — our `"timeout": 3` **is** honored, and is in seconds.
- `discovery.rs:649` — the default when unset is **600s**, and only `SessionEnd` is clamped. Nothing forced 3s on us; it is entirely self-imposed and currently unmeetable.
- `HookHandlerConfig::Command` supports **`"async": bool`** (`hook_config.rs:158`), which maps to `HookExecutionMode::Async` (`engine/mod.rs`). `discovery.rs:500` honors it for **every event except `SessionEnd`**, which is downgraded to sync with a warning.

## Expected

Observation ingress must never sit in the critical path of a tool call.

Suggested shape (all three are independently useful):

1. Declare the pure-ingress handlers `"async": true`. `PreToolUse` currently returns `{}` — it injects no context and has no reason to block. This is the single highest-value change and the host already supports it.
2. Raise `observe_timeout` to something the handler can actually meet on the paths that must stay synchronous because they return `additionalContext` (`PostToolUse`, `Stop`, `SessionStart` return advice today).
3. Stop re-loading the store 17× per invocation — load once per hook process and reuse.

## Guard

There is **no test asserting hook wall-clock stays under the declared timeout** (`grep -rn "observe_timeout" tests/` → no hits; only `test_observe_hooks.py:571` asserts the `drain_preflight_failed` diagnostic string). That is why this regressed silently. Whatever the fix, it should come with a latency assertion against a realistically-sized store, since the cost is store-size-dependent and a fixture-sized store will always pass.

## Why it matters

Release-blocking for v0.1 on the Codex host. This makes yoetz actively harmful to the session it is supposed to help: ~6s added to every tool call, hooks killed mid-write, and — because the killed hook is the thing that drains the outbox — observation stays dark regardless of consent.

Found during dogfood session `019ff71d-4132-7563-98b6-20d403dc94ed` (2026-08-12), which the operator aborted at 75s over this.

### LINKED ISSUE #210: [bug]: Control handshake rebuilds the 69-schema catalog (~1.0s) to hash a 27KB manifest — blows the 0.35s drain preflight in every hook
## Duplicate search

- [x] Searched open issues/PRs for handshake, schema, catalog, drain, preflight, connect, perf. No duplicate. This is the un-fixed second leg of #204 (closed by #208): #208 made the hooks fire; the outbox still never drains, and this is why.

## Area

Control-plane handshake (`service/control_protocol.py`) + schema catalog (`protocol/schemas.py`). Surfaces in the observation hook drain (`cli/observe_hooks.py`).

## Observed — measured 2026-08-12 on this machine, current `main` (`dff08d5`)

`cli/observe_hooks.py:96` sets `_HOOK_CONNECT_PREFLIGHT_SECONDS = 0.35`. A hook that cannot connect within it records `drain_preflight_failed` and returns **without draining anything** (`observe_hooks.py:531-535`).

A hook is a fresh short-lived process on every invocation, so it is **always** a cold connect. Measured cold connect, one per process, five separate processes, service daemon up the whole time:

```
FAILED 10.001s (TimeoutError)
cold connect  5.401s
cold connect  1.009s
cold connect  1.010s
cold connect  1.010s
```

Floor ~1.0s against a 0.35s budget — **~3× over on the best run, and unbounded under contention.** Warm connects in the same process are 0.006s, which is why this is invisible to anything that isn't a one-shot process.

## Root cause

`cProfile` of a single cold `connect_service()`:

```
3.688  service/control_protocol.py:707(client_handshake)
3.670  service/control_protocol.py:693(_manifest_digest)
3.670  protocol/schemas.py:654(load_schema_catalog)
3.548  protocol/schemas.py:520(_plain_schema)                 69 calls
2.892  jsonschema/validators.py:306(check_schema)             69 calls
```

The handshake is:

```python
def _manifest_digest() -> str:
    return load_schema_catalog().manifest_digest
```

and `manifest_digest` is just `sha256(manifest_bytes)` — it depends **only on `manifest.json`**. But `load_schema_catalog()` builds the entire catalog to get there: reads all 69 schema files, canonically re-encodes and byte-compares each one, and runs `Draft202012Validator.check_schema` (JSON Schema *meta*-validation) on each.

Direct comparison, byte-identical result:

```
digest-only (sha256 of manifest.json, 27,373 bytes):   0.0125 ms
full load_schema_catalog():                          1032.9   ms
match: True
```

**~82,000× the necessary work.** `_load_catalog_state` is `@lru_cache(maxsize=1)`, so this is free for long-lived processes and paid in full by every hook, every CLI invocation, and every one-shot client.

Cold connect (1.01s) ≈ catalog load (1.03s). The handshake penalty is entirely this.

## Consequences observed in the live store

`yoetz observe status` for `hmac-sha256:722a4470…2540`:

```
consent: active
lifecycle: stale
undelivered: 73 (cause: mapping_missing=73; last successful drain: 12323.2s ago)
hook_diagnostics: {"count":78,"last_event":"PreToolUse","last_reason":"drain_preflight_failed",
  "reasons":{"consent_revoked":25,"service_unavailable":33,"drain_preflight_failed":15,
             "mapping_missing":4,"drain_budget_exhausted":1}}
gaps: content_capture_unavailable,cursor_stale,mapping_missing,outbox_quarantined,
      service_unavailable,unpaired_event
```

Last successful drain **3.4 hours** ago. During live dogfood session `019ff71d-4132-7563-98b6-20d403dc94ed` the only hook diagnostic emitted was `{"event":"UserPromptSubmit","reason":"drain_preflight_failed","ts":"2026-08-12T17:57:21.578Z"}`.

## Expected

The handshake should obtain the manifest digest without building and meta-validating the catalog. `sha256(manifest.json)` is sufficient and byte-identical; the catalog build can stay lazy for callers that actually need to *validate* something.

Secondarily, `_HOOK_CONNECT_PREFLIGHT_SECONDS = 0.35` should be justified against a measured cold connect rather than a warm one. Even after this fix, 0.35s is a thin margin for a cold process under load — the five samples above spanned 1.0s to >10s, so the tail is contention, not just catalog cost.

## Why it matters

This is the reason observation delivery is still dark after #208. Consent is granted, the policy is trusted, the daemon is up, the hooks now fire — and every drain gives up in the first 0.35s because the client spends its first second meta-validating 69 JSON schemas to hash a 27KB file. It also taxes every ordinary one-shot CLI command by ~1s.

Found during dogfood session `019ff71d-4132-7563-98b6-20d403dc94ed` (2026-08-12).

### LINKED ISSUE #211: [bug]: Observation outbox overflows into an unbounded quarantine (199 entries / 142KB) that is silent, eventually lossy, and taxes every hook
## Duplicate search

- [x] Searched open issues/PRs for outbox, quarantine, eviction, observation, growth, retention. No duplicate. #204 (closed by #208) recorded `quarantine: 0` — quarantine growth is new since that fix landed.

## Area

Observation local store retention: `adapters/integrations/observation_local.py`.

## Observed — live store, read-only inspection 2026-08-12, current `main` (`dff08d5`)

Workspace `/Users/shayb/yoetz-core`, commitment `hmac-sha256:722a4470…2540`. State file is **383,926 bytes** against `_MAX_STATE_BYTES = 1_048_576` (`observation_local.py:66`):

| Field | Count | Bytes |
| --- | --- | --- |
| `quarantine` | **199** | **142,017** |
| `envelopes` | 256 | 160,863 |
| `pending_outbox` | 73 | 57,108 |
| `dedup` | 284 | 21,300 |
| *(everything else)* | — | ~2,600 |

`gaps` includes `outbox_quarantined` and `content_capture_unavailable`.

At the time #204 was filed (2026-08-11) this workspace had `quarantine: 0` and `pending_outbox: 235`. One day later the outbox has been drained-by-overflow into quarantine rather than delivered: **199 observations are now parked in a terminal bucket**, and quarantine is the single largest consumer of the state budget.

## Two distinct problems

**1. Overflow is silent and lossy.** `_save` (`observation_local.py:1385-1425`) enforces the 1 MiB cap by evicting, in order: `envelopes` → `pending_outbox` (moved to `quarantine`) → `quarantine` (evicted outright, via `_record_quarantine_eviction`). So under sustained delivery failure the store first converts undelivered work into quarantined work, then **permanently destroys it**. The only signal is a gap code in a status field nobody is required to read. `quarantine_evicted_count` is currently `0`, so we are not losing data *yet* — but the store is at 37% of cap with the drain broken, and nothing bounds the growth except the eviction path itself.

**2. Retention has no time or count bound.** There is no expiry, compaction, or operator-visible reclaim for `quarantine`. Once an envelope lands there it occupies budget forever. Since the whole file is re-parsed and re-canonicalized on every access, dead quarantine entries are pure tax on every future hook.

## Coupling to hook latency

This is why the two issues filed alongside this one are getting worse over time rather than being constant. Measured canonical-encode cost of the live state, by variant:

```
full (live)                402,577 bytes    35.9 ms
no quarantine              260,562 bytes    22.5 ms
no quarantine+envelopes     99,701 bytes     8.6 ms
empty-ish                   21,262 bytes     1.8 ms
```

`_load` runs **17× per hook invocation** and `_save` re-encodes once per outbox row mutation. Dropping quarantine alone would cut encode cost ~37%. The feedback loop is: drain fails → outbox grows → overflows into quarantine → state file grows → every hook gets slower → hooks exceed their 3s timeout and are killed → drain fails.

## Expected

- Quarantine is bounded by count and/or age, with reclaim, so it cannot become the dominant consumer of a fixed byte budget.
- Overflow eviction is a loud, first-class operator signal, not just a gap code — `yoetz observe status` should say how many observations were quarantined and how many were destroyed, and say it prominently. (`observe status` already reports `undelivered` and `gaps` honestly, which is good; quarantine depth and eviction count are missing from it.)
- A path to drop quarantined entries once the underlying delivery failure is fixed, so a recovered install does not carry the tax forever.

## Why it matters

Silent, unbounded, eventually-destructive accumulation in the subsystem whose entire purpose is producing a trustworthy record. The honesty rules in `CONTRIBUTING.md` make coverage-bounded wording load-bearing; a store that quietly discards observations after quietly failing to deliver them is the version of that we should least tolerate.

Found during dogfood session `019ff71d-4132-7563-98b6-20d403dc94ed` (2026-08-12).

TheGaySupreme123:
Fixes #209. Fixes #210. Fixes #211.

The three issues are one closed loop: a cold control handshake meta-validated all 69 packaged schemas to hash a 27KB manifest (#210), which blew the drain preflight in every hook process, so undelivered work piled into an unbounded quarantine (#211), which grew the state file that every hook re-parsed 17×, pushing every hook past its own declared 3s timeout where Codex SIGKILLed it (#209) — and the killed hook was the thing that drains the outbox.

## Commits, in dependency order

1. **#210 — hash the manifest without building the catalog.** `schema_manifest_digest()` reads and hashes only `manifest.json` (byte-identical result); `load_schema_catalog` stops re-running Draft 2020-12 meta-validation on members whose bytes already matched the packaged manifest digest (81% of measured load cost — that invariant moves to an explicit conformance test over every packaged member). `validate_schema_document` keeps full strict rechecking. Measured: catalog load 954.8ms → 164.8ms; handshake digest 1032.9ms → 0.7ms. Note: a corrupt resource set still fails the handshake — frame validation builds the catalog and raises — just as `frame_invalid` rather than the precise `schema_digest_mismatch`.
2. **#209 — hooks stop blocking tool calls and stop re-parsing state.** Pure-ingress handlers (PreToolUse, PermissionRequest, PreCompact, PostCompact, SubagentStart, SubagentStop) declare `"async": true` — verified against the Codex source at tag `rust-v0.148.0-alpha.6` (the build dogfood runs): async handlers cannot block or return decisions, which these never do (they provably emit `{}` on every path including errors), and unknown handler fields are ignored by older hosts. Advice-returning handlers (SessionStart, PostToolUse, Stop) stay sync at a meetable 10s; SessionEnd keeps 3s because the host hard-clamps it and warns per-session about async SessionEnd. The store gains a bounded, stat-validated parse cache (was 17 re-parses of a ~384KB file per hook); the connect preflight rises to a cold-connect-justified 1.0s.
3. **#211 — quarantine bounded and honest.** Detail expires after 14 days measured from a store-authored `quarantined_at` (state schema `/3`, tolerant in both read directions), fenced on the trusted-clock epoch so a snapshot restore or clock jump pauses the bound instead of firing it. `yoetz observe reclaim --workspace <path>` drops detail explicitly; involuntary evictions and operator reclaims are counted separately (a cleanup never reads as data loss); `observe status` surfaces depth and both counters.
4. **Drain recurrence tax.** Within one pass, a `mapping_missing` rejection retires that session's remaining rows (stamped with the shared cause so `observe status` reports `mapping_missing=N`, never `not_attempted`); workspace-global rejections end the pass; `service_unavailable` stays row-scoped with a consecutive-failure yield. The budget clock starts after the connect. A nonblocking per-workspace drain lease stops Codex's now-concurrent async hooks from re-ingesting the same backlog up to 8×. SessionEnd drains under a 0.75s budget to fit its 3s clamp.

## Guards that did not exist before

`grep -rn "observe_timeout" tests/` used to return nothing — a 3s budget was missed by 63% silently. Now: a wall-clock assertion against a realistic (≥250KB, live-shaped) store, a parse-count guard (== 1; 33 with the cache neutered), an execution-mode contract test that derives the async set from `ADVICE_SAFE_EVENTS` so they cannot drift apart, preflight-vs-budget accounting, poison-row and global-stop pass exits, drain-lease exclusion, the clock-jump fence, and the reclaim/eviction split.

## Review

Three independent adversarial review passes (concurrency/process-lifecycle, protocol/security/host-contract, behavior/backward-compat/test-adequacy) ran against the branch; every confirmed blocker/major was fixed in `2f4ed79` (drain wedge, budget arithmetic, receipt-time vs quarantine-time, unfenced wall clock, unbounded daemon cache, delivery-cause mislabeling, reclaim-as-destruction). The reviews also verified: `_copy_state` covers every mutable field, the bump/ack CAS stays race-safe under real concurrency, at-least-once delivery holds, and the new status fields break no consumer.

## Live verification (this machine, real 384KB store, 80-row dead backlog)

- Hook wall time: 3.06–4.89s (all SIGKILLed) → 1.5–2.5s, against declared budgets of 10s (sync) / 3s (SessionEnd), with PreToolUse & friends async and non-blocking.
- Cold connect: 1.0–>10s → tens of ms.
- End-to-end dogfood (session fixing #206 in a fresh clone, Codex `0.148.0-alpha.6`): every fired hook completed, a ledger task auto-attached, and `observe status` reached `lifecycle: active, undelivered: 0` with drains succeeding continuously — the first fully-working live observation delivery on record. That session's product is PR #212.

## Upgrade note (also in CHANGELOG)

The timeout/async changes alter Codex's per-hook trust hashes: on an existing install the changed hooks report `Modified` and are **silently skipped** until re-approved (Codex's startup hooks review, or re-seeding `[hooks.state]`). Until re-approval only the unchanged hooks fire.

Remaining known non-blockers: the two `test_service_boundary_imports` failures on this machine reproduce on `main` (the >16KiB git-config `state capture` defect, unfiled); the legacy `mapping_missing` backlog rows from dead sessions stay pending by design — that is #205 route-2 territory.

> [!NOTE]
> ### Fix hook latency and session taxing by making observation hooks non-blocking with scoped drain retry logic
> - Selected observation hooks (PreToolUse, PermissionRequest, PreCompact, PostCompact, SubagentStart, SubagentStop) are now declared `async: true` in the Codex hooks manifest so they no longer block tool calls; timeouts for most hooks are raised from 3s to 10s while SessionEnd stays synchronous at 3s.
> - Drain budget now starts after a successful connect; `_HOOK_CONNECT_PREFLIGHT_SECONDS` is raised to 1.0s so a slow but successful cold-connect does not consume the drain budget.
> - Drain passes are now gated by a per-workspace POSIX advisory lock (`drain_lease`) so concurrent hook firings cannot double-drain the same workspace.
> - Retry scoping is introduced: `mapping_missing` retires all remaining rows for that session in the current pass; workspace-global rejections (`vault_locked`, `observation_disabled`, `paused`) end the pass immediately; `service_unavailable` is row-scoped and yields after 3 consecutive occurrences.
> - Quarantine entries now record a store-authored `quarantined_at` timestamp; entries older than 14 days are pruned on save into aggregate evidence; a new `yoetz observe reclaim` CLI command and `reclaim_quarantine` store method allow explicit operator-initiated clearance.
> - In-process state-file and key-material caches are added to `LocalObservationStore` to avoid repeated disk I/O and JSON parsing across hook invocations within the same process.
> - Schema version bumped from `yoetz.observation-local/2` to `/3`; older files are backfilled on load. Risk: callers of `list_quarantine` expecting 3-tuples will break — entries now return 4-tuples including `quarantined_at`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5b494c4.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous observation hooks with expanded execution timeouts.
  * Added `observe reclaim` to remove quarantined observation details.
  * Observation status now reports quarantine depth, eviction counts, and reclaimed counts.

* **Improvements**
  * Improved observation draining, retries, concurrency handling, and quarantine retention.
  * Added schema integrity verification and more efficient schema loading.

* **Documentation**
  * Updated changelog and interface documentation for observation behavior and quarantine management.


TheGaySupreme123:
**Dogfood outcome and the defects it exposed.**

This branch was verified by a live Codex session (`019ff7d2-7923-7103-b0d8-d3944743acdd`) that fixed #206 and opened PR #212. Hook execution is fixed: **77 of 78 hooks completed, zero timeouts, zero kills** across 75 `PostToolUse` invocations, and `observe status` reached `lifecycle: active, undelivered: 0` — the first working end-to-end delivery on record.

That success immediately exposed a successor defect class, now filed as #214–#222. Nothing in it blocks this PR — every one of these was unobservable while the outbox was dark — but #214 should be fixed before observation is relied on:

- **#214 (P0)** `observation_coordinator.py` appends as `mapping.yoetz_writer_id`, the agent's *own* writer, so hook advisories advance the frontier the agent holds. All 3 MCP `publish_work` calls failed (same `request_id`, so the prescribed idempotent replay was genuinely attempted); the agent finished on the CLI.
- **#215–#217** the error vocabulary for that failure contradicts itself; ingest minted 165 duplicate findings that pinned the receipt; ingest also created claims the agent never made, and correctly rejecting one produced two penalty findings.
- **#218–#222** drain throughput ceiling, latched `status.gaps`, and three diagnosability defects.

Recommendation: merge this (the loop it fixes is real and measured), then treat #214 as the next blocker rather than shipping observation as "working".

TheGaySupreme123:
@coderabbit

TheGaySupreme123:
@coderabbit please review the pr

PR #228, opened 2026-08-13
Fix observation ingest dogfood defects (#214-#223)

### LINKED ISSUE #214: [bug]: Observation ingest writes as the agent's own writer, so every cooperative publish_work races it
## Duplicate search

- [x] Searched open issues/PRs for frontier, publish_work, writer, conflict, observation ingest, race. No duplicate. #209/#210/#211 (PR #213) are about hooks *running*; this is about what they do once they run. It could not be observed before PR #213 because no observation had ever been delivered.

## Area

`src/yoetz/application/observation_coordinator.py` (ledger append path) × `src/yoetz/application/publish_work.py` (cooperative optimistic guard).

## Observed — dogfood session 2026-08-12, Codex session `019ff7d2-7923-7103-b0d8-d3944743acdd`, on PR #213

Observation ingest appends to the ledger **as the cooperative agent's own writer**:

```python
# observation_coordinator.py:208, :300, :429
writer_id=mapping.yoetz_writer_id,
```

`mapping.yoetz_writer_id` is the writer minted by the agent's `start` and stored in the lifecycle mapping. So every hook advisory is written by the same writer identity the agent is holding an `expected_frontier` against, and each one advances that writer's frontier between the agent's `status` read and its `publish_work` call.

In a real session the agent could not publish through MCP at all. Three attempts, all with the **same** `request_id` `req_5c4397d9-918b-423b-9289-741090fc9a13` — i.e. the idempotent replay the error message prescribes was actually performed — under task `tsk_56b5be7c-7d1d-49c7-92ca-bafbe33be0ae`, session `ses_734a2ced-9443-49cf-ba4b-7bbe889303db`, writer `wri_73884055-d1d9-49d0-8a18-431161ba1784`:

| time (UTC) | outcome | code | retryable | reason |
| --- | --- | --- | --- | --- |
| 21:14:54 | rejected | `FRONTIER_CONFLICT` | true | `frontier_changed` (`err_1b4f15ba-9c92-4c5c-b6f4-2fe766b7348c`) |
| 21:15:27 | rejected | `FRONTIER_CONFLICT` | true | `frontier_changed` (`err_a604b0a7-5992-416d-b91f-0e1b7f6b5b0a`) |
| 21:15:49 | rejected | `EVENT_INVALID` | **false** | `frontier_changed` (`err_006048f6-8fa2-4085-8dc6-edb968bbc01b`) |

A later CLI attempt hit the same wall (`err_dc59e366-088f-4b96-83c2-e9980a98d5e4`, `FRONTIER_CONFLICT`). The agent abandoned the MCP write surface and completed the session through the local CLI instead.

The third row is the documented escape hatch closing: dropping `expected_frontier` is rejected outright for state-sensitive families (`publish_work.py:593-597`, `_STATE_SENSITIVE_FAMILIES` includes `claim_recorded`, `decision_recorded`, `result_recorded`, `plan_published`). That is a deliberate and correct rule — but it means an agent racing a co-writer has **no** admissible way to publish a plan or a claim: supplying the frontier races, and omitting it is forbidden.

## Expected

Observation ingest is a *second source of evidence*, not a second author of the agent's work. Some separation is needed so hook advisories cannot invalidate a cooperative writer's optimistic guard. Options, in rough order of size:

1. Give observation ingest its own writer identity (harness-authored), distinct from the agent's writer, so the two do not share a writer frontier.
2. If they must share a task, make the optimistic guard writer-scoped rather than task-scoped, so a co-writer's appends do not invalidate the agent's expectation.
3. Failing both, give the cooperative surface an admissible way to publish state-sensitive events under contention (e.g. a compare-and-set on the writer's own head rather than the task frontier).

Whatever the shape: an agent following the documented retry advice exactly, with the same `request_id`, must eventually succeed.

## Why it matters

Release-blocking for v0.1 on the Codex host, and a direct consequence of fixing the delivery loop. Before PR #213 delivery never happened, so the two writers never met; now that it works, **the more observation succeeds, the less usable the cooperative MCP surface becomes**. That inverts the feature: the subsystem meant to raise coverage prevents the agent from recording work at all, and pushes agents onto the CLI where the harness integration earns no `hook_observed` coverage.

Found on PR #213 during the #206 dogfood, which produced PR #212.

### LINKED ISSUE #215: [bug]: EVENT_INVALID for an omitted expected_frontier is mislabeled frontier_changed, marked non-retryable, and prescribes a retry
## Duplicate search

- [x] Searched open issues/PRs for EVENT_INVALID, retryable, frontier_changed, error vocabulary. No duplicate.

## Area

`src/yoetz/application/publish_work.py:416-434` (`_event_invalid`) and `:593-597`.

## Observed — dogfood session 2026-08-12, on PR #213

Omitting `expected_frontier` for a state-sensitive event family raises:

```python
# publish_work.py:593-597
if request.expected_frontier is None and any(
    item.draft.schema in PAYLOAD_TYPES and item.draft.schema.name in _STATE_SENSITIVE_FAMILIES
    for item in drafts
):
    raise _event_invalid("frontier_changed")
```

The live payload (`err_006048f6-8fa2-4085-8dc6-edb968bbc01b`):

```json
{"code":"EVENT_INVALID",
 "message":"The event batch is invalid. Call status to read the current frontier, then retry idempotently with the same request_id.",
 "retryable":false,
 "safe_details":{"reason_code":"frontier_changed"}}
```

Three problems in one response:

1. **The reason code is untrue.** The frontier did not change — the caller omitted it. `frontier_changed` describes a different failure the caller has already seen twice (`FRONTIER_CONFLICT`), so the same reason string now names two distinct conditions with opposite remedies.
2. **`retryable: false` contradicts the message it ships with.** The message prescribes reading `status` and retrying idempotently. The comment at `:424-425` states the intended rule — "only a stale frontier is fixed by rereading status" — which is exactly this case, yet the flag says the call is terminal.
3. **The message does not name the actual fix**, which is "supply `expected_frontier`; it is required for state-sensitive families". An agent that follows the text verbatim retries the identical request and fails identically.

## Expected

A distinct reason code for the omitted-frontier rule (e.g. `expected_frontier_required`), a message naming the field to supply, and a `retryable` flag consistent with that message. If the condition really is terminal for this request shape, the message must stop prescribing a retry.

## Why it matters

`CONTRIBUTING.md` makes error text load-bearing: the remediation contract is that a caller can repair a failure from the response alone. Here the response's three fields disagree with each other, and in the observed session this was the last thing the agent tried before abandoning the MCP write surface entirely (see the frontier-contention issue filed alongside this one).

### LINKED ISSUE #216: [bug]: Observation ingest mints unbounded duplicate findings (165 in one session) that permanently block the receipt
## Duplicate search

- [x] Searched open issues/PRs for findings, duplicate, receipt, unresolved, observation advice. No duplicate. Related but distinct from the frontier-contention issue filed alongside this one: that one is about *writes colliding*, this one is about *what gets written*.

## Area

Observation materialization / advice → findings (`src/yoetz/application/observation_materialize.py`, `observation_coordinator.py`), surfaced through `check` and blocked at `receipt`.

## Observed — dogfood session 2026-08-12, Codex session `019ff7d2-7923-7103-b0d8-d3944743acdd`, on PR #213

Over a single ~19-minute session, observation delivery minted findings faster than any agent could resolve them. `unresolved_finding_count` climbed monotonically:

```
21 → 71 → 141 → 157 → 161 → 165
```

They are essentially one finding repeated: `kind: material_limitation_omitted`, summary "Configured provider is not ready", priority 1. The rollout shows 30 occurrences of that kind in the tool traffic alone. **The agent authored none of them** — they arrive from hook ingest, one per qualifying tool call, with no deduplication against the identical finding already open.

Each unresolved finding pins the receipt at `unresolved_findings_remain`, so the session became structurally un-closeable roughly two minutes in, and stayed that way regardless of how the actual work went.

## Expected

- An advisory finding the source will re-derive on every tool call must be deduplicated against the open finding it repeats, not appended again. Identity should be the (kind, subject, cause) tuple, not the event that observed it.
- A standing environmental condition ("configured provider is not ready") is one fact about the workspace, not N facts about N tool calls.
- Findings the agent did not author should not be able to make a receipt unattainable on their own — or if they can, the receipt must offer a bounded way to dispose of them in one action rather than one-by-one.

## Why it matters

The receipt is the product. A session where the work was completed correctly and the PR was opened (PR #212) could not produce an honest receipt, because a subsystem that is supposed to *raise* coverage generated 165 copies of a single environmental note and blocked it. This also interacts with the drain-throughput issue: every duplicate is another row to deliver and another entry to encode on every subsequent hook.

Found on PR #213 during the #206 dogfood.

### LINKED ISSUE #217: [bug]: Observation ingest creates claims the agent never made, and rejecting them adds penalty findings
## Duplicate search

- [x] Searched open issues/PRs for claim, injected, questionable_finding_rejection, weak_or_stale_response. No duplicate.

## Area

Observation materialization (`claim_recorded` drafts from hook envelopes) × `check` finding logic × `respond` evaluation.

## Observed — dogfood session 2026-08-12, Codex session `019ff7d2-7923-7103-b0d8-d3944743acdd`, on PR #213

Observation delivery created claims the agent never made — `clm_8102520e-ba81-4036-9dab-38b9852b0180` and `clm_ac8a5b41-0de9-489e-b6ee-b27a4a4f0b34` are both present in the session record with no corresponding agent action.

A subsequent `check` then fired `claim_without_admissible_evidence` against one of them — correctly, since a hook-derived claim has no agent-supplied evidence behind it. The agent responded by rejecting the finding and naming the cause (that the claim was injected by observation, not authored by it).

The next `check` added **two more findings against the agent** for that response: `questionable_finding_rejection` and `weak_or_stale_response`. Rollout counts for the session: `claim_without_admissible_evidence` ×3, `questionable_finding_rejection` ×2, `weak_or_stale_response` ×2.

So the cascade is: ingest invents a claim → check flags the claim → agent correctly disowns it → check penalizes the agent for disowning it. There is no move available to the agent that reduces the finding count.

## Expected

1. Observation ingest should not mint `claim_recorded` events on the agent's behalf. A hook observed an action; it did not observe a *claim of completion*, which is an authored assertion with evidence obligations attached.
2. If harness-authored claims are intentional, they must be distinguishable in provenance so `check` does not demand agent-supplied evidence for them.
3. Rejecting a finding that names a real provenance problem must not itself be scored as a questionable rejection. The `respond` evaluation needs to admit "this finding's subject was not authored by me" as a valid disposition.

## Why it matters

This is the honesty machinery turned against a correct answer. The agent did the right thing — refused ownership of an assertion it never made, and said why — and the system's response was to add two more unresolved findings, which in turn contribute to the receipt block described in the duplicate-findings issue filed alongside this one. Any agent that learns from this pattern learns to accept phantom claims rather than dispute them, which is precisely the wrong lesson for a provenance product.

Found on PR #213 during the #206 dogfood.

### LINKED ISSUE #218: [bug]: Hook drain cannot keep pace with a working session — backlog grows and lifecycle degrades to stale
## Duplicate search

- [x] Searched open issues/PRs for drain, throughput, budget, backlog, sweep. No duplicate. #211 bounded quarantine growth; PR #213 made drains succeed. This is the next constraint: successful drains that are simply too small.

## Area

`src/yoetz/cli/observe_hooks.py` drain budget (`_HOOK_DRAIN_BUDGET_SECONDS`, `DEFAULT_OBSERVATION_SWEEP_LIMIT`) and the service-side sweeper.

## Observed — dogfood session 2026-08-12, on PR #213

Delivery worked, then fell behind under ordinary load. Trajectory across the session:

| time | lifecycle | undelivered |
| --- | --- | --- |
| baseline | degraded | 13 |
| 00:15:52 | **active** | **0** |
| 00:17–00:23 | degraded | 2–3 |
| 00:24 → 00:27 | degraded → stale | 8 → 33 |
| session end | stopped | 48 |

`drain_budget_exhausted` climbed monotonically 8 → 68 while drains kept *succeeding* — so this is not the #210 preflight failure returning, it is per-pass capacity. Final workspace state after the session: `undelivered: 41`, all `not_attempted`.

Every hook gets `_HOOK_DRAIN_BUDGET_SECONDS = 1.75` (0.75s on `SessionEnd`) and a `DEFAULT_OBSERVATION_SWEEP_LIMIT` of 64 rows, but a working agent generates events faster than one hook's slice can deliver them, so arrival outruns drain and the backlog grows monotonically once the session gets busy.

## Expected

Delivery should keep pace with an ordinary session, or degrade in a way that is visible and bounded rather than accumulating silently. Candidates: let the service-side sweeper (not the per-hook slice) own bulk delivery so hook drains only handle the newest rows; scale the sweep by observed backlog; or batch multiple envelopes per ingest round trip instead of one row per call.

## Why it matters

Lifecycle reaching `stale` mid-session means coverage claims degrade during exactly the sessions that generate the most evidence. The backlog is also what makes every subsequent hook more expensive, which is the recurrence pattern #209/#211 were filed about — bounded now, but not eliminated.

Found on PR #213 during the #206 dogfood.

### LINKED ISSUE #219: [bug]: observe status gaps latch forever — mapping_missing reported while mapping exists and delivery is clean
## Duplicate search

- [x] Searched open issues/PRs for gaps, mapping_missing, status accuracy. No duplicate.

## Area

`src/yoetz/adapters/integrations/observation_local.py` — `state.gaps` accumulation and `_status_unlocked`.

## Observed — dogfood session 2026-08-12 and after, on PR #213

`yoetz observe status --json` for the dogfood workspace reports:

```json
"gaps": ["content_capture_unavailable","cursor_stale","mapping_missing","outbox_quarantined"],
"mapping_present": true,
"delivery_causes": {"not_attempted": 41}
```

`mapping_missing` is reported while a mapping demonstrably exists, no pending row carries that cause, and delivery had already reached zero undelivered earlier in the same session. `state.gaps` is a `set` that is only ever added to (`note_coverage_gap`), so it accumulates the union of every condition ever encountered and never reflects current state.

The same is true of the other entries: `outbox_quarantined` persists after quarantine is emptied, and `cursor_stale` persists after the cursor advances.

## Expected

`gaps` should describe the workspace's current coverage state, or clearly separate "currently true" from "seen historically". A gap that has been resolved should stop being reported — otherwise the field cannot be used to decide whether observation is healthy right now, which is what both `observe status` and coverage wording use it for.

## Why it matters

Coverage-bounded wording is load-bearing in this project: gap codes flow into what a receipt may claim. A permanently-latched gap either overstates a limitation forever, or trains readers to ignore the field. During this session it also actively misled diagnosis — `mapping_missing` was the headline symptom of #210, so seeing it persist after the fix suggests a regression that had not occurred.

Found on PR #213 during the #206 dogfood.

### LINKED ISSUE #220: [bug]: yoetz version reports internal_error for the diagnosable support_resource_set_mismatch
## Duplicate search

- [x] Searched open issues/PRs for version, internal_error, resource manifest, support_resource_set_mismatch. No duplicate.

## Area

`src/yoetz/version.py:470` (`ResourceIntegrityError("support_resource_set_mismatch")`) surfaced through the CLI error mapping.

## Observed — dogfood session 2026-08-12, on PR #213

Mid-way through a legitimate guidance-regeneration ripple, `yoetz version` failed with:

```
internal_error: the command could not be completed
```

The real cause, visible only in a traceback the agent had to provoke itself:

```
yoetz.version.ResourceIntegrityError: support_resource_set_mismatch
```

which is a precise, expected, and fully recoverable condition: the support resource digest had not yet been regenerated in the documented order. The agent spent several minutes diffing manifests to discover a fact the command already knew.

## Expected

`support_resource_set_mismatch` should reach the operator by name, with the remediation (regenerate the support digest / run the manifest sync) — the same treatment `verify_resource_manifest` already gives its own failures, which print `FAIL (codex_skill_manifest_stale) <path>` and are immediately actionable. `internal_error` should be reserved for conditions the code cannot name.

## Why it matters

`yoetz version` is the command CONTRIBUTING points contributors at to confirm a ripple is complete, so it is most often run in exactly the half-finished state that triggers this. `cli/exits.py` already carries a `REMEDIATION_MESSAGES` map (added in #170) for precisely this class of problem; this path does not use it.

Found on PR #213 during the #206 dogfood.

### LINKED ISSUE #221: [bug]: PluginInspection has no .state and install_plugin refuses a legitimately updated tree without naming the fix
## Duplicate search

- [x] Searched open issues/PRs for PluginInspection, install_plugin, modified_copy. No duplicate.

## Area

`src/yoetz/adapters/integrations/codex_plugin.py` — `PluginInspection` (`:60-77`), `install_plugin` (`:371`).

## Observed — dogfood session 2026-08-12, on PR #213

Two avoidable crashes while regenerating the committed plugin tree:

1. **`AttributeError: 'PluginInspection' object has no attribute 'state'`.** `install_plugin(...)` returns `PluginInspection`, whose fields are `presence`, `trust_observable`, `installed_digest`, `notes` — there is no `state`. The neighbouring activation API (`inspect_activation`, used by `observe status`) *does* return an object with `.state.value`, so `.state` is the natural guess and it fails at runtime rather than at type-check time for anyone scripting the install.

2. **`IntegrationError: modified_copy`** from `install_plugin` against a tree whose managed files had legitimately changed. Recovering requires knowing to pass `replace_modified=True`, which the error name does not suggest and the message does not state.

## Expected

- Either give `PluginInspection` a `state` accessor consistent with `inspect_activation`, or make the mismatch obvious — the two "how is the integration doing?" APIs returning differently-shaped results is the trap here.
- `modified_copy` should say what to pass to proceed, in line with the `REMEDIATION_MESSAGES` convention.

## Why it matters

Minor, but this is the documented path for regenerating the committed plugin tree — the exact operation `tests/packaging/test_committed_plugin_tree.py` tells contributors to perform, and it is performed rarely enough that nobody remembers the flags.

Found on PR #213 during the #206 dogfood.

### LINKED ISSUE #222: [bug]: One Stop hook failed at session teardown while 77 of 78 hooks completed (no diagnosis)
## Duplicate search

- [x] Searched open issues/PRs for Stop hook, hook failed, session end. No duplicate. Distinct from #209: that was every hook exceeding its timeout; here 77 of 78 hooks completed normally and one did not.

## Area

`Stop` lifecycle hook (`yoetz hooks observe --event Stop`), at session teardown.

## Observed — dogfood session 2026-08-12, Codex session `019ff7d2-7923-7103-b0d8-d3944743acdd`, on PR #213

Hook execution across the session was otherwise clean — 77 of 78 fired hooks completed within budget, zero timeouts, zero kills (75 of them `PostToolUse`). The single exception was the final `Stop`:

```
hook: Stop
hook: Stop Failed
```

**Filed without a diagnosis, deliberately.** `Stop` is one of the synchronous advice-returning handlers, it runs at teardown alongside `SessionEnd`, and the session's very last drain recorded `service_unavailable` — but that is a correlation, not a cause, and `handle_observe` is written to exit 0 on every path, so a host-visible "Failed" means either a non-zero exit escaped that contract or the host killed it for another reason.

## Reproduction pointer

Rollout: `~/.codex-testing/sessions/2026/08/13/rollout-2026-08-13T00-13-13-019ff7d2-7923-7103-b0d8-d3944743acdd.jsonl`; stdout log line 8051. Next step is to run a session to completion under a hook wrapper that captures the handler's exit code and stderr, since neither is preserved in the host log.

## Why it matters

Low severity but worth not losing: `Stop` is an advice delivery point, and a teardown-time failure is the least observable kind — the session is over, so nobody is watching. If `handle_observe` can exit non-zero at all, that contradicts its documented "always exits 0" contract and would be worth knowing before v0.1.

Found on PR #213 during the #206 dogfood.

### LINKED ISSUE #223: [bug]: check livelocks under continuous observation traffic
## Defect

`freeze_case` reserves a case at frontier F, while `commit_check_if_current` requires the task head to remain exactly F. Continuous observation delivery can advance the task frontier between those operations. The mismatch is persisted as a terminal `FRONTIER_CONFLICT`, so replaying the same `request_id` cannot resume the operation.

This is the `check` counterpart of #214 and was found while designing the #214–#222 repair.

## Decided scope

Fix in the #214–#222 pass by applying back-pressure to observation-authored appends while a frozen case exists for the session. Return retryable `OPERATION_PENDING` so the observation outbox defers and retries. Do not relax `commit_check_if_current` frontier equality: the verdict must remain bound to the frontier it tested.

## Acceptance

- Observation append during a frozen check case defers with retryable `OPERATION_PENDING`.
- The check can commit at its frozen frontier.
- Deferred observation remains in the outbox and is delivered afterward.
- Cooperative appends and non-observation frontier changes retain their existing behavior.

Related: #214.

## Verified citations (added after reading the code)

Confirmed against `src/yoetz/adapters/memory/ledger.py` on `fix/209-210-211-hook-latency`.

**The reservation binds to frontier F.** `freeze_case` entry compares the caller's expectation to the
live head:

```python
# adapters/memory/ledger.py:1688-1690
projection = self._state.projection
frontier = Frontier(projection.frontier, projection.head_digest)
if expected_frontier is not None and expected_frontier != frontier.sequence:
    raise _frontier_conflict(frontier)
```

and the reservation only commits if the head has not moved while the frozen case was staged
(`finalize` is awaited at `:1745`, so this window *does* contain a real suspension point, unlike the
`append_batch` window):

```python
# adapters/memory/ledger.py:1749-1751
current = Frontier(self._state.projection.frontier, self._state.projection.head_digest)
if current != frontier or self._pending_import(session_id):
    raise _frontier_conflict(current)
```

**The mismatch is persisted as terminal.** `commit_check_if_current` does not merely reject — it
rewrites the operation record to `COMPLETE`/`TERMINAL` and stores the failure:

```python
# adapters/memory/ledger.py:2229-2260 (abridged)
if current != frozen.case.frontier or self._state.frozen_cases.get(key) != frozen.case:
    failure = _frontier_conflict(current)
    ...
    terminal = replace(record, state=OperationState.COMPLETE, phase=CheckPhase.TERMINAL, ...)
    self._state.operations[key] = (terminal, None)
    self._state.check_errors[key] = failure
    raise failure
```

and a replay of the same `request_id` re-raises the stored error rather than resuming
(`:2222 raise stored_error`). That is what makes this a livelock rather than a retryable conflict:
the documented "retry idempotently with the same request_id" remedy is closed off permanently for
that request.

**SQLite inherits this.** `SqliteLedger.append_batch` clones state, builds a `MemoryLedgerAdapter`
over the clone (`adapters/sqlite/repository.py:1358-1370`) and delegates the whole acceptance
decision to it (`:1371`). A fix in the memory adapter is inherited verbatim; parity is structural, so
this needs parity *tests*, not a second implementation.

## The precedent the fix should mirror

`append_batch` already applies exactly this kind of back-pressure for a different in-flight
operation — a `RECEIPT` append defers while an import is pending:

```python
# adapters/memory/ledger.py:1190-1193
if command.operation_kind is OperationKind.RECEIPT and self._pending_import(command.session_id):
    raise _error(PublicErrorCode.OPERATION_PENDING, retryable=True)
```

(`_pending_import` is defined at `:1057-1062`; the same guard appears at `:1301`, `:1638`, `:1681`,
`:1685`, `:1710`, `:1750`, `:2257`.) The proposed gate is the observation-authored analogue, keyed on
`self._state.frozen_cases` (populated `:1751-1780`, cleared at commit).

**Why it degrades cleanly:** `ObservationCoordinator` converts any `PublicOperationError` into
`_reject(SERVICE_UNAVAILABLE)` (`application/observation_coordinator.py:366-382`), which
`route_observation_ingest` classifies as retryable (`application/observation_drain.py:32-43`), so the
row stays in the outbox and is redelivered after the check commits. Nothing is dropped.

TheGaySupreme123:
## Summary

Fixes the observation-ingest defect class exposed by the 2026-08-12 dogfood across #214–#223.

- separates observation materialization onto a deterministic harness writer while retaining safe live-session lookup under the legacy writer
- permits stale cooperative frontiers only when every intervening record satisfies the exact service-stamped observation predicate
- back-pressures observation appends while a check case is frozen, preserving check frontier integrity
- stabilizes candidate-scoped advice identity and stops inferring agent claims from completion hooks
- moves bulk outbox delivery to the leased service sweeper, keeps hook drains small/newest-first, and reports only current observation gaps
- adds actionable version/plugin diagnostics and hardens hook teardown so output failures remain exit-zero and diagnosable
- records the contracts in ADR-022 and updates generated guidance/resource manifests

## Root cause

Successful observation delivery advanced the task-scoped frontier through the cooperative writer. Held `publish_work` and `check` operations therefore raced continuous hook traffic. At the same time, snapshot-scoped advice identities and latched health gaps amplified successful observation into duplicate findings, stale lifecycle state, and blocked receipts.

## Impact

With observation active, cooperative work can retain an expected frontier across observation-only advances. Observation delivery defers during a frozen check instead of invalidating it, standing advice remains bounded, and hook delivery can drain without putting bulk work on the synchronous hook path.

## Validation

- focused adapter/application/CLI/daemon/property/policy suite: 221 passed
- committed resource byte-parity slice: 18 passed
- resource compatibility slice: 10 passed
- Ruff passed
- Pyright: 0 errors, 0 warnings
- resource manifest: 96 resources verified
- `git diff --check` passed

The isolated branch-installed dogfood remains the post-land proof described in the plan.

Closes #214
Closes #215
Closes #216
Closes #217
Closes #218
Closes #219
Closes #220
Closes #221
Closes #222
Closes #223

> [!NOTE]
> ### Fix observation ingest defects to support frontier concurrency and gap lifecycle tracking
> - Adds `has_active_frozen_case` to `LedgerPort` and both memory/SQLite adapters; observation-authored appends now block with `OPERATION_PENDING` while a frozen check holds the session frontier.
> - Introduces `is_observation_authorship`/`is_observation_authored` predicates and a derived `observation_writer_id` so the coordinator consistently uses a deterministic writer identity instead of stored mapping writer IDs.
> - Relaxes `expected_frontier` conflict detection: appends where all intervening records are observation-authored are accepted; missing `expected_frontier` on state-sensitive batches now returns retryable `expected_frontier_required`.
> - Upgrades gap tracking in both memory and local observation stores from a flat set to timestamped, active/inactive `_GapState` records; gaps are resolved when healing signals arrive (cursor advance, content captured, delivery succeeds).
> - Advice materialization now ties findings to explicit subject event references, deduplicates via projection, and derives `FindingId` from the advice-candidate digest rather than the evidence-basis digest; schema bumped to `obs-advice/1.1.0`.
> - Risk: `LocalObservationStore` schema bumps from `/3` to `/4`; `MATERIALIZATION_MAPPING_VERSION` changes to `obs-ledger/1.2.0`; advice finding IDs will change for existing standing conditions.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 03407c2.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Improved observation tracking with active gap status and first/last-seen timestamps.
  - Added safer handling for concurrent observation processing and frontier updates.
  - Finding identities are now stable per observation candidate.
  - Added clearer integration and resource-installation status reporting.

- **Bug Fixes**
  - State-sensitive publications now require `expected_frontier` and return a retryable error when omitted.
  - Prevented duplicate observation processing and unauthorized authorship bypasses.
  - Improved recovery from output failures, resource inconsistencies, and interrupted observation delivery.

- **Documentation**
  - Clarified observation claims, publication rules, frontier conflicts, and lifecycle evidence handling.


TheGaySupreme123:
@macroscopeapp please review this pr

TheGaySupreme123:
@macroscope please review this pr

TheGaySupreme123:
@coderabbit please review

Issue #232, opened 2026-08-13
[bug]: one corrupt pending check makes the observation sweeper saturate the service and starve all MCP/control calls


TheGaySupreme123:
## Duplicate search

- [x] Searched open and closed issues/PRs for `storage_corrupt`, `observation_ingest_ledger_append_storage_corrupt`, `observation_ingest_advice_storage_corrupt`, service sweeper starvation, high CPU, and MCP `start` hangs.
- #218 is related but not a duplicate: it addressed drain throughput/backlog and led to the service-side bulk sweeper. This issue is the poison-record failure mode introduced/exposed by that solution: retryable failures consume the service until unrelated control traffic cannot run.
- #214 is not a duplicate: it addressed observation writer/frontier races, not logical resume corruption or service starvation.
- #210 is not a duplicate: it removed full schema-catalog construction from the handshake. The measured loop here is ledger recovery during observation ingest after the handshake path was fixed.
- #24 is not a duplicate: it addressed a stop/drain deadlock after an accepted control response, not READY maintenance monopolizing the event loop.
- #229-#231 are unrelated to this failure.

## Area

- `src/yoetz/application/observation_drain.py::ObservationOutboxSweeper.sweep`
- `src/yoetz/service/daemon.py::_run_ready_maintenance`
- `src/yoetz/application/observation_coordinator.py::ObservationCoordinator.ingest_request`
- `src/yoetz/adapters/sqlite/repository.py:548` while reconstructing pending frozen cases

## Observed — 2026-08-13, exact current `main` after PR #228

Repository/runtime:

- `main` / `origin/main`: `9ccc08258c9ff69feea4bde288245fda9de8f400` (merge PR #228)
- isolated Codex task: `019ffb1b-48c2-7ce1-8b55-c4e351569b36`
- MCP request: `start`, request id `req_8df04731-c2c8-4c96-8f07-4ca8230b154e`

The first real `mcp__yoetz__start` call remained pending for more than two minutes. Replaying the exact same idempotent request also remained pending. The user eventually interrupted the task.

This was not a model/tool-authoring problem:

- Yoetz initialized and served its six real MCP tools and guidance resources.
- The request passed model-side validation and entered the actual `mcp__yoetz__start` call.
- The agent correctly preserved the same request id instead of creating duplicate task state.
- It never reached repository inspection or editing because `start` never returned.

Live process evidence while the call was pending:

```
yoetz service run   CPU 79% -> 91% -> 100%
[...]

Issue #229, opened 2026-08-13
[bug]: state capture is unavailable on any repo whose .git/config exceeds 16 KiB, bounded by a stderr constant


TheGaySupreme123:
## Duplicate search

- [x] Searched open and closed issues/PRs for `state capture`, `read_limit_exceeded`, `subject state`, git config size. No duplicate. Related in *class* to #220 (opaque error surfacing) but a different defect: this one is the bound itself.

## Area

`src/yoetz/adapters/git_subject_state.py:711` (the `.git/config` read) and `:44` (`_STDERR_LIMIT`).

## Observed — 2026-08-13, on `main`

`yoetz state capture` is entirely unavailable on this repository, and two packaging tests fail on `main` because of it:

```
tests/packaging/test_service_boundary_imports.py::test_state_capture_lazily_imports_only_subject_state_modules
tests/packaging/test_service_boundary_imports.py::test_state_capture_command_performs_no_write_or_network_effect
```

Root cause, with the exception surfaced by calling the adapter directly (the CLI hides it — see the companion issue):

```
yoetz.adapters.git_subject_state._CaptureFailure: read_limit_exceeded
  File "src/yoetz/adapters/git_subject_state.py", line 920, in _read_small_mutable
  File "src/yoetz/adapters/git_subject_state.py", line 257, in open_local_workspace
ValueError: read_limit_exceeded
```

The read at `:711`:

```python
config = git_root / "config"
try:
    encoded = _read_small_mutable(config, _STDERR_LIMIT)
```

Two problems in one line:

1. **The cap is the wrong constant.** `_STDERR_LIMIT` (`:44`, `16_384`) is named and sized for captured subprocess stderr. Reusing it as the `.git/config` size bound means a future change to stderr handling silently changes which repositories can be captured at all. Every other bound in this module has a purpose-named constant (`_COMMAND_OUTPUT_LIMIT`, `_PATH_OUTPUT_LIMIT`, `_READ_CHUNK`).

2. **16 KiB is too small for an ordinary working repository.** This repo's own `.git/config` is **16,749 bytes** — 115 sections, of which **112 are `[branch ...]`**. Nothing exotic: the file grows roughly linearly with the number of local branches, so any long-lived clone crosses the bound eventually. There is no warning as it approaches, and no signal that branch count is what did it.

[...]

PR #245, opened 2026-08-13
Fix the 2026-08-13 dogfood defect batch (#235, #237-#243)

### LINKED ISSUE #235: [bug]: first MCP call after cold start returns SERVICE_UNAVAILABLE — daemon listens before it can handshake, and the on-demand connector abandons its 30s budget on the daemon it just spawned
## Duplicate search

- [x] Searched open and closed issues/PRs for `SERVICE_UNAVAILABLE`, cold start, handshake, unresponsive, `connect_service_on_demand`, accept loop.
- #233 (closed by PR #234) is the parent, not a duplicate: it added the end-to-end connect/handshake bound and the accepted-but-unresponsive fail-fast. This issue is a residual gap in that fix: the fail-fast also fires on the legitimately starting daemon the same call just spawned.
- #232 (closed) owned steady-state starvation by the observation sweeper. This failure occurred on the post-#234 build and self-resolved within ~20 seconds; it is a startup-window defect, not steady-state starvation.
- #229–#231 are hook/state-capture/receipt defects; none touch the connect path.

## Area

- `src/yoetz/service/daemon.py::ServiceDaemon.start` / `serve` — endpoint published and listening before the accept loop exists
- `src/yoetz/service/client.py::connect_service_on_demand` — `_AcceptedServiceUnresponsive` fail-fast abandons the 30s budget
- `src/yoetz/mcp/server.py::ensure_service_client` — a single aborted connect surfaces directly to the model

## Observed — 2026-08-13, isolated Codex session `019ffb69-22d9-74e1-bb2e-a7c33cbca8a0`

Build: post-PR #234 working tree (editable install; the daemon's log origins moved from `repository.py:548` to `:572` across the restart, confirming the new build was running).

- ~13:58:54.7Z — first real `mcp__yoetz__start`, request id `req_6a3d84d8-f45f-4a68-87ea-882601c37324`. No daemon was running (last old-daemon log line 13:44:41Z).
- 13:58:54 local (`ps lstart`) — the failing call itself spawned the daemon (`python3 -m yoetz service run`, pid 50744, started `Thu Aug 13 16:58:54 2026`).
- 13:59:02.848Z, after 8.1s wall — the call returned:

  ```text
  SERVICE_UNAVAILABLE
  retryable: yes
  The local service is unavailable; retry after it is ready.
  correlation: err_8043f24b-d4fd-4a71-a7d0-ad33e99c616a
  ```

  The durable diagnostic line shows the id was bridge-minted, i.e. the daemon never answered:

  ```json
  {"timestamp":"2026-08-13T13:59:02.848Z","correlation_id":"err_8043f24b-d4fd-4a71-a7d0-ad33e99c616a","component":"mcp.bridge","operation":"start_public_error","reason":"service_unavailable","request_id":"req_6a3d84d8-f45f-4a68-87ea-882601c37324"}
  ```

- 13:59:13.5Z — the new daemon's first log line (observation coordinator): it needed roughly 19 seconds from spawn to serving.
- ~13:59:30Z — an identical same-request-id retry succeeded in 11.3s (`tsk_6c924b55-115c-47d9-b824-8dfd4cba45ec`). The service was healthy; only the startup window was mishandled.

8.1 seconds is far under the nominal 30-second on-demand budget (`_SERVICE_START_TIMEOUT_SECONDS`, `client.py:126`), so the connector did not exhaust its budget — it aborted.

## Root-cause chain

Two halves; either alone would have let the first call succeed.

1. **Server: the daemon accepts connections long before it can answer a handshake.**
   `ServiceDaemon.start` publishes the control endpoint first (`daemon.py:570` `lifecycle.publish_endpoint()`), and `bind_control_listener` calls `listen()` at bind time (`adapters/control/unix_socket.py:432`), so from that moment every client `connect()` succeeds via the kernel backlog. `start()` then runs `activate_ready_application` (`daemon.py:589`) — vault, catalog open, pending frozen-case reconstruction — before returning, and only after that does `serve()` create the `_accept_loop` task (`daemon.py:692-694`). In this run that window was ~19 seconds. For its whole duration a connected client's handshake is never read.

2. **Client: `connect_service_on_demand` treats accepted-but-unresponsive as terminal even for the daemon it just spawned.**
   PR #234 (`c9b6f1e`) introduced `_AcceptedServiceUnresponsive` (`client.py:1100-1104`): a socket that is accepted but whose `client_handshake` does not complete within `_CONNECT_HANDSHAKE_TIMEOUT_SECONDS` = 5s (`client.py:125`) raises it, and `connect_service_on_demand` re-raises it immediately both before spawn (`client.py:1256-1259`) and inside the post-spawn poll loop (`client.py:1275-1276`), abandoning the remaining budget. The rationale — a wedged pre-existing owner cannot be repaired by a successor — is correct before spawn, but in the post-spawn loop the unresponsive process is the one this very call started milliseconds earlier, and continuing to poll until the deadline is exactly the right remedy; no successor is involved.

   Observed arithmetic fits: spawn at ~13:58:54.8 → a few seconds of connection-refused polls until `publish_endpoint` bound the socket → one accepted connection → 5s handshake bound → raise at 13:59:02.8. Total 8.1s of a 30s budget, with the daemon becoming healthy ~11s later.

3. **Bridge: connection failures are not retried.** `_invoke_with_reconnect` only reconnects around `invoke`; `ensure_service_client` sits outside its `try` (`mcp/server.py:509-523`), so the single aborted attempt surfaced to the model on the documented first-call path.

Regression window: PR #234 merged 2026-08-13T13:51:28Z; this failure occurred at 13:59:02Z on the first cold start after the merge.

## Expected

- The documented cold-start path (`connect_service_on_demand` spawning the daemon) uses its full budget before surfacing SERVICE_UNAVAILABLE: after this call spawned the daemon, accepted-but-unresponsive within the budget means "still starting", not "wedged".
- Either the daemon does not accept control connections until it can answer handshakes (publish/listen after activation, or answer with a typed starting/locked handshake state during activation), or the client tolerates the startup window, or both.
- The pre-spawn fail-fast for a genuinely pre-existing unresponsive daemon (the #232 incident shape) is preserved.
- The end-to-end bound from #233 is preserved — no return to unbounded waits.

## Acceptance criteria

- [ ] Subprocess test: with no daemon running and an activation that takes longer than the handshake bound, the first MCP `start` succeeds within the on-demand budget instead of returning SERVICE_UNAVAILABLE.
- [ ] The pre-existing wedged-daemon case still fails fast without spawning a successor (regression cover for the #232/#233 behavior).
- [ ] The total connect attempt stays within the 30-second bound in both cases.

## Why it matters

The very first tool call of every fresh session on a cold machine is the one most likely to hit this window, and it presents Yoetz as unavailable at the exact moment the agent decides whether to adopt it. This agent retried and recovered, but the failure cost a retry loop and ~40 seconds on a healthy install, and a less persistent agent would simply proceed without Yoetz.

## Design gate

Probably not: an ordering/containment fix within existing lifecycle semantics; no protocol, privacy, or durability change. If the remedy changes when the endpoint file appears (observable by the CLI), call that out in the PR.

Found during the 2026-08-13 dogfood session `019ffb69-22d9-74e1-bb2e-a7c33cbca8a0` (~16:57–17:00 local / 13:57–14:00Z).

### LINKED ISSUE #237: [bug]: yoetz service run collapses service_already_running into internal_error exit 70, misreporting a live daemon as a failed restart
## Observed

During dogfood codex session `019ffb69-22d9-74e1-bb2e-a7c33cbca8a0` (2026-08-13 14:22:42Z), the agent — after `service status` reported `service_unavailable` and prescribed `run 'yoetz service run' under your selected user supervisor` — ran exactly that and got:

```
internal_error: the command could not be completed
```

exit code 70, in 0.47s. The agent (reasonably) concluded the service was dead and unrestartable, abandoned the completion publish/check/receipt sequence, and ended the session with no receipt.

## Root cause

The daemon (pid 50744) was alive the whole time — merely unresponsive on its control plane (companion starvation issue). The restart chain was:

1. `service_run()` (`src/yoetz/cli/app.py:906`) → `ServiceDaemon.serve()` → `lifecycle.acquire_singleton()` (`src/yoetz/service/daemon.py:568`).
2. The live daemon still held the singleton lock → `src/yoetz/service/lifecycle.py:286` raises `LifecycleError("service_already_running")` — the CORRECT refusal.
3. No handler between there and the CLI entry point understands `LifecycleError`; it lands in the blanket `except Exception` at `src/yoetz/cli/app.py:2107-2109`, which prints `internal_error: the command could not be completed` and exits 70.

So the truthful message "a service is already running (it may be unresponsive)" was collapsed into a generic internal error — sending the whole incident down the wrong diagnostic path. Worse, the remediation loop is self-contradictory: `service status` says the service is unavailable and tells you to run `service run`; `service run` correctly refuses because the daemon exists; the user/agent is left with two commands that each point at the other.

## Expected

- `yoetz service run` against a live-but-held singleton reports `service_already_running` (with the holder pid) and a distinct exit code — not `internal_error`/70.
- `service status`'s `service_unavailable` remediation should distinguish "no daemon (start one)" from "daemon exists but is unresponsive" (surface the lock holder; suggest inspecting/waiting/superseding rather than a start that must refuse).
- The CLI catch-all at `cli/app.py:2107-2109` should map known `LifecycleError` reasons to their bounded remediation strings before degrading to `internal_error`.

## Impact

P1 for dogfood ergonomics: this single misreport converted a transient starvation outage into "the service died and cannot be restarted" in the session narrative, and blocked any recovery attempt the agent might have reasoned toward (e.g. waiting and retrying status, which would in fact have succeeded — the daemon recovered by ~14:30Z).

## Refs

Session rollout lines 460-472 (local file); `src/yoetz/service/lifecycle.py:286`; `src/yoetz/cli/app.py:906, 2107-2109`. Related: #232/#234 (starvation class), #235 (cold-start window), companion starvation-recurrence issue.

### LINKED ISSUE #238: [bug]: post-#234 daemon still starves: control plane unreachable ~10-20 min with zero diagnostics, then silent recovery
## Observed

During dogfood codex session `019ffb69-22d9-74e1-bb2e-a7c33cbca8a0` on 2026-08-13, the local service (pid 50744, `svc_690db93e-c003-4630-a55e-368527caaae8`, spawned 13:58:54Z by the session's first `start`) became progressively unreachable and stayed that way for the session's entire completion phase, then recovered on its own without ever logging the outage or the recovery:

- 14:02:31Z / 14:03:00Z — last confirmed healthy `publish_work` round trips.
- ~14:06–14:09Z — hook drains that passed preflight connect started exhausting their budgets (`drain_budget_exhausted` in hook-diagnostics.jsonl; requires a successful preflight, `observe_hooks.py:577-583`), i.e. the service was already degrading.
- 14:12:57Z — bridge envelope-recovery status RPC failed in ~5s (`err_94380925`, surfaced as `OPERATION_PENDING`/`operation_recovery_unavailable`).
- 14:13:10Z — `status` → `SERVICE_UNAVAILABLE` `err_0af85991` (connect-failure branch, `mcp/server.py:469-477`).
- 14:22:21Z — `status` retry → `SERVICE_UNAVAILABLE` `err_d89886d9`; fresh-process `yoetz service status --json` agreed (`service_unavailable`).
- ~14:30Z+ — the SAME instance (same pid, same service id, generation 51) answered `service status` with `state:"ready"`, `state_reason:"none"`.

The daemon process never exited (no crash report, continuously alive), and `service.diagnostics.jsonl` / `service.stderr.jsonl` contain **zero daemon-side lines between 13:59:13Z and after recovery** — its only log line ever was `observation_ingest_ledger_append_storage_corrupt` `err_50ec3353` (origin `yoetz.adapters.sqlite.repository:572`) at 13:59:13Z. A live `sample` after recovery shows a healthy idle loop, but 11m39s cumulative CPU and a 347 MB peak footprint (118 MB current) indicate a sustained burst saturated the single event loop. The workspace observation outbox (`observation/workspaces/722a4470….json`, 510 KB) was churning throughout, and hooks reported `drain_lease_contended`/`drain_budget_exhausted` on essentially every event for the whole session.

## Impact

This ate the entire integrity payoff of the session: the completion `publish_work`, `check`, and `receipt` could never run. The agent finished honestly with "No Yoetz completion verdict or receipt exists for this run" — the tool's core deliverable was unavailable exactly when it mattered. Combined with #237, the agent was also led to believe the service was dead and unrecoverable.

## Expected

1. Control RPCs stay serviceable while ingest/sweeper work runs (this is the #232 defect class recurring on the post-PR-#234 build — HEAD `99c8b1e` includes c9b6f1e/75aff37, so the starvation fix is insufficient under a corrupt-pending + large-outbox workload).
2. A daemon that cannot service its control plane for minutes MUST emit diagnostics about it (saturation entry/exit, queue depths, loop lag). The zero-log outage makes post-hoc root-causing impossible — the precise saturating loop could not be pinned for this incident.

## Refs

- Related: #232 (closed by PR #234 — this is a recurrence of the class, not a reopen-identical), #233, #235 (same session's cold-start window), #237 (the internal_error misreport that compounded this outage), PR #236 (the session's work product).
- Evidence: `~/Library/Logs/yoetz/service.diagnostics.jsonl` lines 934-939, `~/Library/Application Support/yoetz/observation/hook-diagnostics.jsonl`, codex rollout lines 411-472 (local file).

### LINKED ISSUE #239: [bug]: bridge-side publish_work validation failure (even dry_run) escalates into OPERATION_PENDING when the recovery oracle is unreachable, burying the known INVALID_REQUEST
## Duplicate search

- [x] Searched open and closed issues for `OPERATION_PENDING`, `operation recovery`, `dry_run`, `recovery unavailable`. No duplicate. #65 ("Recovery must not mask a known authoring error") is the closed ancestor of exactly this failure mode and is linked below; #46/#36 shaped the recovery design; #235 covers the cold-start SERVICE_UNAVAILABLE and #232/#233 the starvation that made the oracle unreachable — this issue is about what the bridge *returns* when that happens.

## Area

`src/yoetz/mcp/server.py:871-886` (bridge-side validation failure → recovery escalation), `:700-726` (`_publish_recovery_unavailable_result` assembly), `:131-136` (`_OPERATION_RECOVERY_UNAVAILABLE_MESSAGE`). Note: the string `dry_run` does not appear anywhere in `src/yoetz/mcp/server.py` — the recovery path never consults it.

## Observed — 2026-08-13 dogfood, session 019ffb69-22d9-74e1-bb2e-a7c33cbca8a0

At **14:12:52Z**, `publish_work` with `dry_run: true` (request_id `req_c53f5e9e-…`, `expected_frontier` sequence 41) returned after exactly **5015 ms**:

> Error OPERATION_PENDING; retryable: yes; correlation: `err_94380925-6e66-4b7c-819d-ed0c1ff2e395`
>
> "Operation recovery could not determine whether this request_id already names a committed operation. Retry with the same request_id. If recovery later reports the operation absent, correct the named authoring fields and resubmit with the intended request identity. Hint: authority admits values matching `^[A-Za-z0-9._:-]{1,128}$`; see yoetz://guidance/request-templates.md for a complete decision_recorded request"

`safe_details`: `{field: "/event_drafts/0/payload/authority", reason_code: "operation_recovery_unavailable"}`.

What actually happened (established by forensics on the rollout): the draft genuinely carried an invalid `authority` value in its `decision_recorded` payload and failed **bridge-side** Pydantic validation at `server.py:871-886` — the request never reached the daemon. The bridge then ran its envelope-recovery status lookup (`_publish_recovery_from_envelope`, `server.py:729+`) against a daemon that was unreachable (starved — the mid-session service death is being filed separately), got `UNAVAILABLE`, and `server.py:886` promoted the whole thing to retryable `OPERATION_PENDING`.

Evidence transcript is the codex rollout for this session (local file), not attached.

## The defect

A request that failed **local** validation deterministically never committed *on this submission* — and under `dry_run: true` it could never commit at all, so "could not determine whether this request_id already names a committed operation" is a question that has no reason to be asked, let alone answered with "unknown". Concretely:

1. **`dry_run` is never consulted.** The recovery escalation at `server.py:882` runs for every publish validation failure. For a dry run the recovery oracle's only legitimate purpose (detecting a prior commit under the same request_id) cannot change the correct answer: an invalid body dry-runs to INVALID_REQUEST regardless of any prior operation.
2. **The composite error mixes two unrelated failures and buries both.** The real input defect (invalid `authority`) is demoted to a conditional afterthought ("if recovery later reports the operation absent…") and a hint, while the real availability failure (daemon unreachable) is visible only as `reason_code: operation_recovery_unavailable` inside `safe_details`. Nothing says "your request failed validation locally" and nothing says "the service is down".
3. **The prescribed remedy cannot succeed.** "Retry with the same request_id" resubmits the same invalid body; each retry burns the ~5 s recovery deadline and returns the same ambiguity. The agent followed the hint's *other* branch instead and went chasing the authority regex — which happened to be right, but only by luck of which fragment it trusted; the error's own primary instruction was a dead end.

The design comment at `server.py:688-693` ("Only FOUND replaces the authoring diagnostic. ABSENT yields the original field-pointed validation result. UNAVAILABLE is ambiguity-safe") shows the intent: ambiguity-safety for the *durability* claim. But it achieves that by replacing a certain, actionable INVALID_REQUEST with an uncertain OPERATION_PENDING — trading a correct answer the bridge already holds for a maybe. #65's rule — recovery must not mask a known authoring error — is violated on exactly the UNAVAILABLE leg.

## Expected

- `dry_run: true` + bridge-side validation failure → plain INVALID_REQUEST with the field-pointed locations, no recovery lookup at all.
- Non-dry-run + validation failure + recovery UNAVAILABLE → the validation result should still surface as the primary diagnostic; the durability caveat ("a prior operation under this request_id could not be checked; the service was unreachable") belongs alongside it, not in place of it.
- The message must state the two facts it knows for certain: this body failed validation locally, and the service could not be reached — instead of a conditional that asserts neither.

## Severity

**High.** This fires on the most common failure pair in real sessions (authoring mistake × service unhealthy), converts a deterministic one-shot correction into a retry loop with a ~5 s tax per attempt, and its prescribed remedy is guaranteed not to work.

### LINKED ISSUE #240: [bug]: unknown payload keys yield a hint that recites already-satisfied requirements and never says "unknown property"
## Duplicate search

- [x] Searched open and closed issues for `additionalProperties`, `unknown property`, `hint`, `extra_forbidden`, `invalid_type_or_value`. No open duplicate. Closed ancestors: #48 (authorable event batches), #73 (safe corrective fields), #93 (actionable rejections), #129 (self-correcting nested validation failures) — each improved this hint machinery; none covers the unknown-payload-key case, which the machinery still cannot name.

## Area

`src/yoetz/mcp/errors.py:271-360` (`authoring_hint`), `:396-434` (`_event_draft_hint_parts`), `:437-455` (`_is_event_draft_index_pointer` — deliberately matches `/event_drafts/N/payload`), `:244-268` (`safe_validation_locations`, pointer projected "to the nearest allowlisted parent so untrusted leaf keys (payload fields, extras) never appear"), `:224-241` (`_reason_from_validation_item` — collapses to `invalid_type_or_value`), `:133-149` (`_SAFE_VALIDATION_REASONS`, which *does* contain `"extra_forbidden": "extra_forbidden"`).

## Observed — 2026-08-13 dogfood, session 019ffb69-22d9-74e1-bb2e-a7c33cbca8a0

At **14:01:55Z** a `publish_work` dry_run failed INVALID_REQUEST (correlation `err_cfd7a127-89bb-4f66-8359-1af4bf61f8dc`, **11.1 s** to return) because two **unknown** payload keys — `requested_items` and `source_refs` — were present on `obligation_published` payloads. The response said:

> "each event_drafts entry requires event_id, schema, occurred_at, causal_parents, payload, artifact_refs, and evidence_refs; schema.name admits action_recorded, …"

Every listed required key **was** present and `schema.name` **was** an admitted value. `safe_details` said only `{fields: ["/event_drafts/1/payload"], reasons: ["invalid_type_or_value"]}`.

Evidence transcript is the codex rollout for this session (local file), not attached.

## The defect

The validator never says "unknown/additional property" and never names (or even counts) the offending keys, while the hint affirmatively recites requirements that were already satisfied. The agent's cheapest reading of this error is "one of these required keys or the schema name is wrong" — i.e., the hint steers *away* from the actual defect.

Mechanism, from the code:

1. The unknown leaf keys are not in `_SAFE_LOCATION_SEGMENTS`, so `safe_validation_locations` projects the pointer up to `/event_drafts/1/payload` (`errors.py:258-262`) — correct for privacy, but it discards the only locator the caller needed.
2. The reason token arrives as `invalid_type_or_value`, not `extra_forbidden`, even though the closed vocabulary supports `extra_forbidden` (`errors.py:136`). The payload union's unknown-key failure is collapsed by the fallback in `_reason_from_validation_item` (`errors.py:224-241`), so even the *class* of the mistake is erased.
3. `_is_event_draft_index_pointer` deliberately matches `/event_drafts/N/payload` (`errors.py:446-455` — its docstring says the generic loop "can name nothing there"), so `_event_draft_hint_parts` fires and emits the envelope-requirements + schema-family recital (`errors.py:427-433`) — a hint written for the "cannot author an envelope at all" failure, attached to a failure that is nothing like it.

Privacy does not require this. The key *names* may be untrusted, but the *fact* "the payload carries N properties this family does not admit" is schema-derived and safe. (Whether the names themselves are safe to echo is a separate call — pydantic knows them, and a bounded/escaped echo is worth considering — but the class of defect must be stated either way.)

Separately: **11.1 s** to produce a deterministic, bridge-local validation verdict. Per `server.py:871-886` a validation failure now always runs the envelope-recovery status lookup against the service before the INVALID_REQUEST can surface (the ABSENT leg), so every authoring mistake is taxed with a service round-trip — see the companion OPERATION_PENDING issue for what happens when that lookup can't complete at all.

## Expected

- `extra_forbidden` survives to `safe_details.reasons` for unknown payload keys (the vocabulary already has it).
- The hint for that reason says what happened: "the payload carries properties the `<family>` schema does not admit; admitted keys are …" — admitted keys are frozen schema content, exactly what `authoring_hint` is allowed to speak.
- The requirements recital is reserved for locations whose reasons are `missing`/type-level, so a hint never asserts facts the request already satisfies.
- A locally decidable validation failure should not need a service round-trip before it can be reported.

## Severity

**Medium-high.** Unknown-key mistakes are among the most common authoring errors an agent makes, this is the one case where the hint machinery actively misleads rather than merely under-informs, and it cost this session a multi-minute detour plus 11.1 s per attempt.

### LINKED ISSUE #241: [bug]: standing provider_not_ready advice delivered 29 times byte-identical in one session — suppression identity churns with every envelope while the rendered text never changes
## Duplicate search

- [x] Searched open and closed issues for `advice`, `storm`, `provider_not_ready`, `reissue`, `suppression`, `injection`. No duplicate. #216 (closed) bounded duplicate *finding materialization in the ledger*; #231 (open) and the ADR-022 amendment decide whether `provider_not_ready` becomes a finding at all. This issue is the third leg neither touches: **delivery cadence of the hook-channel advice text itself**. #211 (closed) taxed hooks via the outbox; different mechanism.

## Area

`src/yoetz/application/observation_advice.py:264-271` (`_suppression_identity` = hash of finding ids + **evidence basis digest** + next action), `:366-373` (basis digest computed over the candidates **and every retained envelope**), `:281-306` (`should_reissue_advice`), `:459-471` (rebuild gate); `src/yoetz/adapters/integrations/observation_local.py:783-812` (`peek_advice_for_delivery` — dedupes on `suppression_identity`); `src/yoetz/cli/observe_hooks.py:985-987` (advice refreshed on every hook), `:1091-1098` (delivered as `additionalContext` on every `ADVICE_SAFE_EVENTS` hook), `:474-491`-region (`hook_advice_context` — the rendered text).

## Observed — 2026-08-13 dogfood, session 019ffb69-22d9-74e1-bb2e-a7c33cbca8a0

Between **13:59:32Z and 14:23:14Z** the codex hook channel injected this developer message **29 times, byte-identical** (6 in the first 5 minutes, 23 more after):

> "Yoetz: Configured provider is not ready. Reason: Semantic or provider binding is configured but not ready. Next: connect_provider. Evidence: fireworks."

It was attached to unrelated `rg` / `git status` / `pytest` outputs, and it is not actionable by the agent — `connect_provider` is a human/installation action. Each copy consumed context in a turn where it changed nothing.

Evidence transcript is the codex rollout for this session (local file), not attached.

## Root cause — the dedup key churns while the rendered text cannot

The suppression machinery exists and *looks* correct in isolation: `peek_advice_for_delivery` returns a snapshot only when its `suppression_identity` differs from the last delivered one (`observation_local.py:800-810`), and `should_reissue_advice` keeps the prior snapshot when identities match (`observation_advice.py:293-294`).

But `_suppression_identity` includes `evidence_basis_digest`, and that digest is computed over **all retained envelopes** (`observation_advice.py:366-373`). Every routine tool call ingests a new envelope, so for a *standing* condition like `provider_not_ready`:

- the condition, ranked findings' rule code, next action, and the rendered `hook_advice_context` text are all **unchanged**, yet
- the basis digest — and therefore the suppression identity — changes on **every hook**, so the "once per suppression identity" gate opens on essentially every `PostToolUse`.

The rendered text (`hook_advice_context`) contains none of the changing material (no digest, no frontier), which is why 29 deliveries were byte-identical. So this is not `should_reissue_advice` being "broken" line-by-line — it is the identity it keys on being wrong for standing machine conditions: evidence-sensitivity is the right trigger for advice *about the work*, and the wrong one for advice about a static environment fact whose evidence basis is "everything observed so far".

## Expected

- Delivery dedup keyed on what the agent will actually receive (rendered content / condition identity: rule code + next action + evidence ref), not on a digest over the whole envelope stream.
- Standing machine conditions the agent cannot act on (`connect_provider` and kin) get a bounded cadence — e.g., once at session start and once at Stop/SessionEnd — rather than a per-tool-call channel.
- Distinct from #231: even after the ADR-022 amendment stops materializing this as a finding, the advice channel as coded would still redeliver it every hook.

## Severity

**Medium.** No data corruption, but it consumed context on nearly every turn of a real session (29 injections in 24 minutes), trained the agent to ignore the advice channel, and the affected condition (no provider configured) is the default state for a new user.

### LINKED ISSUE #242: [bug]: synchronous observe hooks add 3.5–7.5s to every workspace exec call (~60–70% of session wall clock) after the #209 timeout-loop fix
## Duplicate search

- [x] Searched open and closed issues for `latency`, `hook`, `slow`, `PostToolUse`, `timeout`. No open duplicate. Closed neighbours: #209/#210/#211 were the hook **timeout loop** (hooks exceeding their own 3 s timeout, catalog rebuild, quarantine tax), fixed by PR #213 — this session confirms the loop is gone (no timeout kills observed). What remains, and was never separately filed, is the **standing synchronous per-call latency** of the observe hooks themselves.

## Area

`src/yoetz/adapters/integrations/codex_plugin.py:183-190` (PreToolUse observe declared `"async": true`), `:110-113` ("Codex ignores unknown handler fields, so older hosts that predate `async` parse this unchanged and **simply run the handler sync**"), `:192-213` (PostToolUse observe **synchronous by design** — it returns `additionalContext`), `:121-127` (execution-mode split rationale from #209); `src/yoetz/cli/observe_hooks.py:87-109` (`_HOOK_DRAIN_BUDGET_SECONDS = 0.20`, `_HOOK_CONNECT_PREFLIGHT_SECONDS = 1.0`, "a hook is always a fresh process"), `:579` (cold-connect preflight), `:741` (start-correlation RPC, `deadline_ms=5_000`), `:985-987` (per-hook advice refresh).

## Observed — 2026-08-13 dogfood, session 019ffb69-22d9-74e1-bb2e-a7c33cbca8a0

Throughout the session, shell commands whose self-reported inner wall time was 0-3 ms reached the model **3.5-7.5 s** later. Representative (measurement method: the command's own `wall_time_seconds` vs the outer "Wall time" in the codex exec output):

- `date -u` — inner ~0 ms, outer **6.7 s** (14:01:21Z)
- `git status --short --branch` — outer **6.3 s**
- `rg` — outer **6.8 s**
- Two commands batched under one `Promise.all` at ~14:22Z: inner 0.0000026 s + 0.0000034 s, outer **10.1 s** — i.e., the tax is paid **per command**, so batching does not amortize it.

The overhead is warm-up sensitive: ~7.5 s early in the session, ~3.5 s late. Roughly **60-70 % of total session wall clock** was this overhead.

Evidence transcript is the codex rollout for this session (local file), not attached.

## Root-cause pointers

The #209 fix moved ingest-only hooks to `"async": true`, but two synchronous paths still sit in every exec call's critical path:

1. **PostToolUse observe is synchronous by design** (`codex_plugin.py:192-213`) because it is the advice `additionalContext` channel — so every tool call waits for a full hook execution even when there is no new advice (which, per the companion advice-storm issue, there almost never is).
2. **PreToolUse's `async: true` is advisory**: on hosts that predate the field it is silently ignored and the hook runs sync (`codex_plugin.py:110-113`). On such a host both hooks serialize into every call.

Each hook execution is a **fresh `yoetz` process** (`observe_hooks.py` comment at :103-106). Measured on this checkout: `python -c "from yoetz.cli import observe_hooks"` costs **0.44 s warm** (bare interpreter: 0.01 s) — cold, with linker/FS caches empty, considerably more, which matches the 7.5 s→3.5 s warm-up curve. On top of interpreter+import each pass can spend up to 1.0 s connect preflight (`:579`), 0.20 s drain budget, per-hook advice recompute over all retained envelopes (`:985-987`), and on start-correlation paths a 5 s RPC deadline (`:741`). Two hooks × (imports + preflight + drain + advice) lands squarely in the observed 3.5-7.5 s band.

The budgets bound each *internal* stage, but nothing bounds — or even measures — the end-to-end synchronous cost per tool call, and process startup (the dominant term) is outside every budget.

## Expected

- A per-tool-call synchronous budget for the whole observe path, with process-startup included — or removal of process startup from the path entirely (persistent lightweight forwarder / daemon-side tap, lazy imports so the hook entrypoint doesn't pay the full CLI import graph).
- PostToolUse advice delivery decoupled from ingest: ingest can be async; the sync leg should return immediately when the advice suppression gate has nothing new (today it does the full refresh first).
- Detection of the sync-downgrade: if the host ignored `async: true`, the plugin should know and say so once, because the latency profile it produces is 2x.

## Severity

**High.** Not a correctness failure, but 60-70 % of a real session's wall clock is pure harness overhead on every single workspace exec call — it dominates the cost of using yoetz under codex and compounds with every other defect that needs retries.

### LINKED ISSUE #243: [bug]: state capture counts .git internals and gitignored trees against the 10,000-entry bound — yoetz cannot capture its own repository, and the limitation never names the bound
## Duplicate search

- [x] Searched open and closed issues for `file_limit`, `state capture`, `subject state`, `unsupported`. No duplicate. #229 (open, fix in PR #236) is the `.git/config` 16 KiB bound — this is the *next* bound the same command trips once #229 is cleared. #230 (open) is the CLI mislabeling structural failures; related in class, different defect: this one is the bound itself, and its unlabeled report.

## Area

`src/yoetz/adapters/git_subject_state.py:627-668` (`_reject_unsafe_tree_entries` — the walk that trips), specifically `:641-645` (`entries_seen > self._max_files` → `FILE_LIMIT_EXCEEDED`) and `:665-667` (the root `.git` directory is *descended into*: the rejection applies only when `directory != workspace.root`, and every directory is appended to `pending`); `:570-573` (`_hash_untracked`'s separate FILE_LIMIT check, which correctly uses `ls-files --others --exclude-standard`); `src/yoetz/ports/subject_state.py:25` (`MAX_SUBJECT_STATE_FILES = 10_000`); call site `git_subject_state.py:444`.

## Observed — 2026-08-13 dogfood, session 019ffb69-22d9-74e1-bb2e-a7c33cbca8a0

At **14:21:45Z**, after the #229 fix on the PR #236 branch unblocked the oversized `.git/config`:

```
yoetz state capture --workspace /Users/shayb/yoetz-core --json
→ {"status":"unsupported","limitations":["file_limit_exceeded"],"files_hashed":0,...}
```

So the tool cannot produce a subject state for its own development checkout, and every dogfood session runs without state capture.

Evidence transcript is the codex rollout for this session (local file), not attached.

## Root cause — the bound is applied to the wrong population, and the report doesn't say which bound

`files_hashed: 0` shows the failure happens before any hashing: it is `_reject_unsafe_tree_entries`, which counts **every filesystem entry under the root** against `MAX_SUBJECT_STATE_FILES = 10_000` — including:

- **gitignored trees**: `.gitignore` is never consulted; this checkout's `.venv` alone contributes the bulk of ~30,757 non-`.git` entries;
- **the entire `.git` object store**: the `.git`-rejection clause at `:665` exempts the *root* `.git`, and the unconditional `pending.append` then walks it (7,275 entries here).

Measured on this checkout: **38,032 total entries**, versus **917 tracked files**. The population the capture actually hashes is bounded elsewhere and correctly (untracked hashing uses `ls-files --others --exclude-standard`, `:563`, with its own 10,000 check at `:570`), so the safety walk is 40x more restrictive than the thing it guards, and inconsistent with it.

Two defects:

1. **Wrong population / too-small default.** Bounds are legitimate — the walk exists to keep the safety scan cheap. But counting `.venv` and `.git` internals against a 10,000-entry ceiling makes any medium-sized real repository (anything with a vendored env, `node_modules`, or a long-lived clone) structurally uncapturable, while a fresh CI clone passes. Same shape as #229: the defect reaches users and never the test suite.
2. **The limitation is unlabeled.** `limitations:["file_limit_exceeded"]` does not say *which* of the three FILE_LIMIT sites tripped (`:301`, `:572`, `:643`), what the count was, or what the bound is — so the operator cannot distinguish "too many working files" from "your virtualenv is inside the repo" from "your .git is large", and cannot tell how far over they are.

## Expected

- The unsafe-tree walk skips (or bounds separately, and generously) `.git` internals and gitignore-excluded trees, so its count approximates the population capture actually touches — or the ceiling is raised to something a real repository fits under.
- On any limit trip, the limitation names the walk/bound that tripped and reports observed count vs limit (both are integers already in hand; neither discloses content).

## Severity

**Medium-high.** `state capture` backs ADR-011 structural subject-state (release gate **E-015**), and today it cannot run on the project's own repository — so all dogfood evidence is collected without subject state, on exactly the class of repository (real, lived-in) the feature exists for.

TheGaySupreme123:
Fixes the eight-defect batch from the 2026-08-13 dogfood session `019ffb69` — the cold-start window, the misreported restart, the silent starvation, the two bridge-error defects, the advice storm, the hook tax, and the uncapturable repository.

Closes #235. Closes #237. Closes #238. Closes #239. Closes #240. Closes #241. Closes #242. Closes #243.

## What changed, per issue

**#235 — cold start returned SERVICE_UNAVAILABLE.** The daemon now publishes its control endpoint only after activation settles (READY or LOCKED), so a connect during startup is refused — which `connect_service_on_demand` already treats as "still starting" — instead of accepted-and-never-answered. The client's post-spawn poll loop now keeps polling an accepted-but-silent daemon it just spawned to the 30 s deadline; the pre-spawn fail-fast for a genuinely wedged pre-existing owner (#232/#233 shape) is byte-identical. A subprocess test injects a 7 s activation delay and proves the first on-demand connect succeeds inside the budget (it reproduces the incident's exact `service_unavailable` failure on `main`).

**#237 — `service run` collapsed `service_already_running` into `internal_error`/70.** Known `LifecycleError` reasons now map to truthful bounded messages (`service_already_running` → `BUNDLE_BUSY`, exit 20, with the holder pid when probeable from the advisory stamp in the already-held lock file); `invalid_transition` alone remains `internal_error`/70. `service status` distinguishes "no daemon — start one" from "a daemon is listening but not answering — `service run` will refuse; use `service stop`". No exit-code or `PublicErrorCode` additions; everything stays inside the pinned sets.

**#238 — silent control-plane starvation.** Three mechanisms, all addressed: (1) the observation sweeper ran every store call — each a blocking cross-process `flock` plus a full workspace-document re-encode — on the event-loop thread; sweeper and coordinator store calls now run off-loop (the sweeper on its own bounded 4-thread executor so a stranded worker can never exhaust the shared pool), which also makes the 30 s sweep deadline enforceable for the first time. (2) A corrupt pending row made every later RPC repeat the entire failed ledger recovery (O(ledger) per call, the incident's 347 MB / 11 CPU-minutes); a deterministically failed recovery now latches its verdict — but only when the causal chain contains no `OSError`, so a transient EMFILE/EIO cannot brick an intact bundle. (3) A watchdog OS thread samples event-loop lag and appends `control_plane_saturation_entered`/`_persists`/`_cleared` (with `duration_ms`/`operation_count`) directly to the durable diagnostics ring — deliberately bypassing the stdlib stderr logger so a full stderr pipe cannot block it — so the next outage is diagnosed while it happens.

**#239 — validation failure escalated to OPERATION_PENDING.** A declared `dry_run: true` (exact boolean) that fails bridge-local validation returns the field-pointed `INVALID_REQUEST` with no recovery lookup at all — a dry run appends nothing, so no prior-operation lookup can change the answer. For real submissions the FOUND leg is preserved (a committed same-`request_id` operation still wins), and the unavailable leg now keeps the validation result primary: non-retryable `INVALID_REQUEST` stating both known facts (the body failed locally; the prior-operation check could not run), with `reason_code: operation_recovery_unavailable` beside the `fields`/`reasons` locations.

**#240 — unknown payload keys misdiagnosed.** The filed hypothesis (pydantic union collapse) was wrong: drafts are validated by jsonschema, and the closed `extra_forbidden` token was destroyed at the jsonschema→pydantic boundary while the oneOf scorer weighed discriminator failures from branches the caller never selected — actively pointing at `schema/name`, a field that was correct. The scorer now restricts itself to the branch the caller's discriminator selected, `extra_forbidden` survives to `safe_details.reasons`, and the hint names the admitted keys (frozen schema content, version-matched so a multi-version family like `evidence_recorded` can never be shown the wrong version's key list) and a bounded count — never the caller-controlled key names. Frozen payload property names joined `_SAFE_LOCATION_SEGMENTS` behind a new import-time completeness gate, so the session's actual shape (a bad value under `source_refs`) now points at that key instead of collapsing to the payload root with an irrelevant envelope recital.

**#241 — 29× byte-identical advice.** Hook delivery is deduplicated on a content identity over the stable condition (rule code, next action, summary, detail) — deliberately excluding evidence refs, which for rules like `observation_gap_or_stale` are a rolling window over the envelope stream and would recreate the churn. `suppression_identity` and its three other consumers (ledger history uniqueness, materialization retry-safety, `observe status` frontier) are untouched. Standing machine conditions the agent cannot act on (`connect_provider`) travel only on session-boundary events and fall through to the next actionable item on per-tool-call hooks. The delivery identity is committed only after the stdout write succeeds, so a failed emit redelivers rather than permanently suppressing.

**#242 — 3.5–7.5 s hook tax.** Two dominant terms measured and removed: the import graph (~325 ms/hook → ~45 ms via a minimal entry shim, a PEP-562-lazy `yoetz.application`, and a leaf `hook_io` module — `hooks.json` unchanged) and, larger on a lived-in store, 10–18 full serialize+fsync cycles of the workspace state per hook (→ a write batch flushed once before any service RPC plus one batch per drain row, an unchanged advice snapshot no longer rewritten, and the `enqueue_outbox` triple-encode collapsed). End-to-end timing is recorded with a `hook_budget_exceeded` diagnostic and a subprocess wall-clock guard — the only test that can see process startup. Measured: ~0.55 s per hook against a realistic store, from 1.7–2.5 s.

**#243 — repository uncapturable.** The pre-hash safety walk no longer descends into the root `.git` (its threat surface is independently covered by `_verify_git_metadata` and the hardened git runner; nested-`.git` rejection and every per-entry check are preserved outside gitignore-excluded subtrees) and prunes fully-ignored subtrees using one `git ls-files --others --ignored --exclude-standard --directory` call — git's own exclusion semantics, so a partially-tracked directory is never pruned. Any file-count limit failure now reports which bound tripped with observed count and limit (`limit_detail`, integers only).

## Adversarial review

Three independent reviewers attacked the combined diff (concurrency/ordering, injection/privacy, semantics). Every confirmed finding was fixed on this branch; the notable ones:

- The recovery latch originally treated a transient `OSError` as permanent corruption (P1 — fixed with a causal-chain check).
- The first content identity for advice still churned for `observation_gap_or_stale` (its evidence ref is a rolling window; P1 — fixed by excluding evidence refs), and the identity was persisted before the emit (P1 — fixed with commit-after-emit).
- The admitted-keys hint could show the wrong version's key list for `evidence_recorded` (P1 — fixed with version-matched branch selection).
- Two designed features were **removed rather than repaired** after review proved them net-negative: the async-downgrade detector (its predicate flags fully compliant hosts — `post_start ≥ pre_finish` is true on async hosts too, and a false claim about the user's environment violates the honesty rules; hook-side detection is not implementable without the tool's own start time) and the advice relevance gate + sidecar (inert in exactly the gap-y workspaces #242 targets, while carrying a real withheld-advice defect).

## Behavior changes to note

- `control.sock` appears **after** activation (ADR-008 amended). Anything waiting on the endpoint file as a readiness signal waits longer — and no longer gets a socket that hangs. During the (now connection-refused) startup window, each concurrent on-demand bridge may spawn one doomed daemon that exits 20 immediately; bounded and self-correcting.
- `service run` against a live daemon: exit 70 → exit 20.
- A dry-run replay of a committed `request_id` no longer surfaces `REQUEST_IDENTITY_CONFLICT` — a preview makes no durability claim.
- `#240` moves pointers: unknown key → `/event_drafts/N/payload` + `extra_forbidden`; bad value under a payload key → that key's pointer.
- A nested `.git` inside a *fully gitignored* subtree is no longer rejected by capture (deliberate, pinned by a test: capture never opens files there).
- Standing advice can reappear at most once per Codex turn (`Stop` fires per turn), and only if different advice was delivered in between.
- `state capture` runs 8 git invocations (was 6); the ignored-prefix call measured ~34 ms.
- The persisted advice-suppression slots change value domain (`suppress-…` → `deliver-…`); no migration needed in either direction (first post-upgrade hook delivers once, correctly).

## Verification

- Full suite (unit, integration, conformance, property, capability, subprocess): **all green** on the assembled branch; ruff + ruff format + pyright strict clean.
- Every headline regression test was verified to **fail on `main`** (including subprocess reproductions of the incident's exact `service_unavailable` and exit-70 shapes, and the 29×-delivery storm).
- `tests/packaging` note: two state-capture packaging tests fail **on the author's machine only** because this checkout's `.git/config` exceeds the 16 KiB bound — that is issue #229, owned by open PR #236, whose check runs strictly before the walk this PR fixes. They pass on a fresh clone (CI). The packaging test now also asserts `status == "captured"`, making it the literal #243 regression guard once #236 lands. `git merge-tree` confirms this branch and #236 auto-merge cleanly with no semantic conflict.

## Known residuals (deliberately out of scope)

- An extra key on the draft's `schema` object (not the payload) still misdiagnoses to `/schema/name` — pre-existing, structurally symmetric oneOf survivors, documented in-code.
- Control RPCs can still *transitively* wait on a worker parked in the store's unbounded flock; the loop no longer parks directly, and the watchdog now reports it if it bites.
- While verifying, a new pre-existing defect was found: the advertised MCP `publish_work` `input_schema` drops the `evidence_recorded` 1.1.0 draft branch and rejects two of its own worked examples (`_event_family_from_draft_branch` only recognizes `-1.0.0` refs) — filed separately.

https://claude.ai/code/session_01MVmv7cnAc3yA6u7X43DJic

> [!NOTE]
> ### Fix dogfood defect batch: harden lifecycle errors, observation sweeps, advice delivery, and MCP validation
> - **Service lifecycle**: the control endpoint is now published only after the accept loop is armed; a second `service run` against a held singleton exits with code 20 and a `service_already_running` message instead of an internal error; the singleton lock file records the holder PID so CLI messages can include it.
> - **On-demand connector**: `connect_service_on_demand` no longer fails fast on `_AcceptedServiceUnresponsive`; it polls until the deadline, tolerating a daemon that accepted but hasn't completed its handshake yet.
> - **Observation sweeps**: sweep operations are offloaded to a thread pool to avoid blocking the event loop; sweeps are bounded by a deadline with distinct `sweep_deadline_exceeded` / `sweep_failed` diagnostics; the first sweep now waits for the control accept loop to arm.
> - **Control-plane watchdog**: a new `ControlPlaneWatchdog` in [`loop_health.py`](https://github.com/TheGaySupreme123/yoetz/pull/245/files#diff-4be50159ebcd66adc1b1610b1838f478e829e777d5f11f82a856d67baadbe320) samples event-loop lag from a background thread and emits bounded diagnostics on saturation entry, persistence, and clearance.
> - **Advice delivery**: `LocalObservationStore.refresh_advice` returns a structured `AdviceDelivery` with item selection, delivery identity, and rendered text; advice is committed under a lease; store mutations within a hook pass are coalesced via a new `batched` context manager to prevent durable acknowledgements ahead of ingest.
> - **MCP validation**: unknown payload keys now yield `extra_forbidden` locations annotated with family, version, and bounded count; authoring hints list admitted keys; recovery-unavailable results return `INVALID_REQUEST` (non-retryable) instead of `OPERATION_PENDING`; dry runs skip recovery lookup entirely.
> - **CLI entry fast path**: `yoetz hooks observe` dispatches through a new [`entry.py`](https://github.com/TheGaySupreme123/yoetz/pull/245/files#diff-8fba2947b707940878f97054fbe29fc5aa81d7279f1418b30e55c7ddc83fe512) fast path that avoids importing the full CLI, reducing startup latency; exceptions in the fast path emit an empty JSON object and exit 0.
> - **Store lock timeout**: interprocess and intra-thread store lock acquisition is now bounded and raises `TimeoutError('observation_store_lock_timeout')` instead of blocking indefinitely.
> - **git subject state**: directory traversal skips `.gitignore`-excluded subtrees entirely; limit closures include structured `SubjectStateLimitDetail` with bound, observed, and limit values.
> - Risk: `refresh_advice` signature changed to return `AdviceDelivery | None`; callers must call `commit_advice_delivery` separately to persist suppression state.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2c00ecd.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Improved publish validation with clearer field-specific guidance and safer recovery handling.
  * Added more truthful service startup, lifecycle, and singleton-lock diagnostics.
  * Added detailed file-limit reporting for state capture.
  * Improved hook performance, advice deduplication, and delivery tracking.
  * Added bounded timing and operation-count diagnostics.

* **Bug Fixes**
  * Fixed cold-start connection handling and control-plane responsiveness.
  * Improved recovery behavior for storage failures and dry-run validation.
  * Excluded ignored files and directories from state-safety checks.
  * Improved event-draft and schema validation messages.


TheGaySupreme123 (commented):
I reviewed the full PR diff and current head CI, and checked the existing CodeRabbit threads to avoid duplicates. These inline notes are additional findings only. The required `unit-and-conformance` job is currently failing; the other required jobs and Security and Privacy completed successfully.

_Done by ChatGPT._

src/yoetz/application/observation_advice.py
```diff
@@ -471,18 +477,92 @@ def build_observation_advice_snapshot(
...
+    rule code, next action, summary and detail are all in the key.
+    """
+
+    top = item
+    if top is None and snapshot.ranked_items:
+        # Mirror hook_advice_context: an item-less call renders the top item.
+        top = snapshot.ranked_items[0]
+    condition: JsonValue
+    if top is not None:
+        condition = {
```
TheGaySupreme123, line 541: _🎯 Functional Correctness_ | _🟠 Major_

**Do not collapse distinct advice conditions that share one rendered template.**

`summary` and `detail` are fixed per rule, so this identity drops the data that distinguishes actual conditions. For example, `failed_command_unresolved` stores the correlation key only in `detail_token`, `subagent_finding_unaddressed` stores the subagent ID there, and `provider_not_ready` stores the missing provider set in `evidence_refs`. After one such item is committed, a later distinct item with the same rule, action, and template hashes to the same `deliver-*` value, so `peek_advice_for_delivery()` suppresses it.

Keep rolling envelope references out of the hash, but carry a rule-specific stable condition identity from `ObservationAdviceCandidate` into `AdviceItem` and hash that. Add a regression covering command A being delivered and resolved, followed by command B failing.

_Done by ChatGPT._

src/yoetz/application/observation_coordinator.py
```diff
@@ -194,6 +195,18 @@ class ObservationCoordinator:
...
+        """Run one blocking local-store call off the service event loop.
+
+        Every ``LocalObservationStore`` entry point takes the same unbounded cross-process
+        flock, so a read is no safer on the loop than a write: whoever holds the lock decides
+        how long the daemon stops answering control clients (#238). Reads route through here
+        too. The store's own reentrant thread lock is taken and released inside a single worker
+        thread on each hop, so no lock is ever held across an await.
+        """
+
+        return await asyncio.to_thread(call)
```
TheGaySupreme123: _🩺 Stability & Availability_ | _🟠 Major_

**Do not send cancellable flock waits to the shared default executor.**

Cancelling `asyncio.to_thread()` does not stop a worker already blocked in the store's unbounded cross-process flock. The 30-second maintenance deadline can therefore return while this call continues, and each later sweep can strand another default-executor worker. This defeats the isolation described in `ObservationOutboxSweeper`: ledger replay and other coordinator calls use the same pool and can eventually be starved.

Use a coordinator-owned bounded executor with an in-flight fence, or make lock acquisition bounded so a timed-out call cannot outlive its request.

_Done by ChatGPT._

src/yoetz/application/observation_drain.py
```diff
@@ -99,13 +123,50 @@ class ObservationOutboxSweeper:
...
+            )
+            self._executor = executor
+        return loop.run_in_executor(executor, call)
+
+    def close(self) -> None:
+        """Release the sweeper's worker pool without waiting on a parked flock."""
+
+        executor, self._executor = self._executor, None
+        if executor is not None:
+            executor.shutdown(wait=False, cancel_futures=True)
```
TheGaySupreme123, line 165: _🩺 Stability & Availability_ | _🟠 Major_

**`wait=False` does not make a blocked worker exit-safe.**

`ThreadPoolExecutor.shutdown(wait=False)` returns immediately, but running worker threads remain alive and Python joins them before interpreter exit. `cancel_futures=True` only cancels queued calls. If a store call is parked in the cross-process flock, service teardown can return while the process remains alive, and that call can resume and mutate local state after the ready application has closed.

Bound or interrupt lock acquisition, or move this work to an execution primitive whose blocked workers cannot hold process exit. Add a subprocess test that closes with one worker intentionally parked and verifies the process terminates.

_Done by ChatGPT._

tests/unit/cli/test_observe_hooks.py
```diff
@@ -1114,7 +1116,319 @@ async def connect(_kind: object):
...
+    )
+    assert code == 0
+    delivered = "additionalContext" in out.getvalue().decode()
+    # Exact accounting, so a regression cannot hide inside a loose ceiling:
+    #   1 local-pass batch flush
+    # + 1 per drained outbox row, bounded by _HOOK_DRAIN_ROW_LIMIT (4 here)
+    # + 1 advice-delivery commit, and only when advice actually reached stdout.
+    # Seventeen were measured before the write batch. Nothing else writes: the
+    # advice sidecar and the async-pair sample are gone.
+    assert _suffix_counts(written) == {".json": 5 + int(delivered)}, written
```
TheGaySupreme123, line 1202: _🧪 Testing_ | _🟠 Major_ | _⚡ Quick win_

**Do not assert four acknowledgements under a wall-clock drain budget.**

Production stops when the 200 ms budget expires, before it necessarily reaches the four-row cap. The current required CI run completed only three acknowledgements, so this exact count failed with four `.json` writes even though 2,524 other tests passed. Runner speed now decides whether the required job is green.

Make the drain clock deterministic for this test, or assert `1 + acknowledged + int(delivered)` using an explicitly controlled acknowledgement count instead of hard-coding four drained rows.

_Done by ChatGPT._

src/yoetz/protocol/schemas.py
```diff
@@ -1014,23 +1078,136 @@ def score(error: ValidationError, path: tuple[str | int, ...]) -> int:
...
+    instance = error.instance
+    schema = error.schema
+    if not isinstance(instance, Mapping) or not isinstance(schema, Mapping):
+        return 0
+    properties = cast(Mapping[str, JsonValue], schema).get("properties")
+    declared: set[object] = (
+        set(cast(Mapping[object, object], properties)) if isinstance(properties, Mapping) else set()
+    )
+    unknown = {key for key in cast(Mapping[object, object], instance) if key not in declared}
+    return min(len(unknown), MAX_UNKNOWN_PROPERTY_COUNT)
```
TheGaySupreme123: _🎯 Functional Correctness_ | _🟡 Minor_ | _⚡ Quick win_

**Do not render a saturated count as an exact cardinality.**

For any payload with more than 32 unknown properties, this returns 32. `authoring_hint()` later says that the payload carries 32 properties, which is false for 33 or more. The request-size bound still permits that shape.

Return an overflow marker, render `at least 32 properties`, or drop the number once the cap is reached. Add a 33-key regression case.

_Done by ChatGPT._

PR #263, opened 2026-08-14
Enforce receipt coverage capacity

### LINKED ISSUE #262: [bug]: retained coverage unions above 64 have no honest receipt representation
## Reporter note

This issue was researched and written by **OpenAI Codex** while dispositioning the unresolved review thread on PR #260.

## Duplicate search

- [x] Searched open and closed issues and PRs for `known_gaps`, `invalid_known_gap`, receipt coverage, overflow, aggregation, and truncation.
- #259 covers the concrete failure where one retained historical observation gap is omitted after recovery. It does not define behavior when the exact retained union is larger than the receipt schema can represent.
- #253 covers missing finding citations during case construction; #34 covers check applicability. Neither addresses coverage-set cardinality.
- PR #260 correctly exposes the retained coverage fold required by #259, but is not the right scope for changing the coverage algebra or wire bounds.
- No duplicate issue or PR was found.

## Area

Coverage algebra and receipt construction:

- `src/yoetz/protocol/coverage.py` — `Coverage` and `weakest`
- `src/yoetz/application/receipt.py` — current-case/check/retained-finding fold
- `src/yoetz/domain/receipts.py` — exact top-level/finding coverage invariant
- `schemas/common/coverage-1.0.0.schema.json`
- `schemas/receipts/receipt-document-1.0.0.schema.json`
- `docs/INTERFACES.md`

## Reproduction

Two coverage values can each be independently valid while their exact union is not representable:

```python
left = Coverage(..., known_gaps=tuple(f"historical_gap_{n:02d}" for n in range(32)))
right = Coverage(..., known_gaps=tuple(f"historical_gap_{n:02d}" for n in range(32, 65)))

weakest(left, right)
# ProtocolValueError("invalid_known_gap")
```

Verified locally on PR #260:

```text
left_valid=32 right_valid=33 union=65
weakest_error=invalid_known_gap
```

Each value satisfies the public coverage schema's per-value `maxItems: 64`. A `Finding` owns one such independently valid `Coverage`, and a receipt may carry many retained findings. There is no task-global admission invariant requiring the union across retained finding rows to stay within 64.

At receipt time, `application.receipt._context` folds the current deterministic case, applicable check, and each retained finding with `weakest`. The 65th distinct code raises before a receipt can be built. Because `ProtocolValueError` is a `ValueError`, the inner receipt-construction handler re-raises the unclassified reason and the outer handler reports:

```text
INVALID_REQUEST
The receipt request is invalid.
```

The request is not malformed, and retrying cannot repair the retained task history.

## Contract conflict

All of the current rules are locally reasonable but cannot hold together for this state:

1. `Coverage.known_gaps` is exact, sorted, unique, and capped at 64.
2. `weakest` unions gaps and explicitly refuses to strengthen or drop representable gaps.
3. Receipt top-level coverage must equal the weakest fold of every carried finding.
4. Every explicit receipt-gap code must occur in top-level coverage.
5. Receipt gaps and coverage gaps are both capped at 64.
6. Current authority says overflow fails case freezing rather than truncating.

Therefore a local “truncate before `weakest`” repair would be dishonest and incomplete: it would silently discard material limitation identities, and `ReceiptDocument.__post_init__` / `receipt_weakest_coverage` would fold the carried finding coverages again.

Catching the exception as `STORAGE_CORRUPT` would improve classification but still leave a valid durable task with no honest receipt representation.

## Reachability assessment

- Current first-party observation producers emit a closed `ObservationGapCode` enum of 20 codes, so the ordinary #259 path (`cursor_stale`) does not by itself approach 64.
- The durable `Coverage` and observation envelope validators nevertheless accept any schema-valid gap token, up to 64 per value.
- Deterministic, semantic, publication, availability, check, and observation paths contribute additional gap families.
- Retained findings can originate at different historical frontiers, so each can be valid while their task-level union grows across time.

This makes immediate probability lower than the #259 regression, but impact high and durable once reached: every receipt attempt fails non-retryably, and no caller action can remove the historical facts safely.

## Why this needs its own design-gated issue

PR #260 restores the existing exact-weakest contract for ordinary retained findings. Fixing this overflow requires choosing a different system-wide invariant, with schema, fixture, codec, receipt-validation, and documentation consequences. It should not be smuggled into #259 as a local aggregation tweak.

Possible design directions include:

1. **Canonical bounded overflow representation.** Define an explicit, deterministic overflow fact (potentially count/digest-bound) and update `weakest`, receipt validation, schemas, and renderers so bounded aggregation is honest rather than silent truncation.
2. **Task-global admission invariant.** Prove before accepting any finding/material coverage that the retained union remains representable, with an explicit policy for the finding that would cross the limit. Simply dropping that finding is not acceptable.
3. **Closed globally bounded gap vocabulary.** Register every gap code and enforce a system-wide vocabulary whose complete union cannot exceed the receipt bound. This constrains compatibility and future extensions.
4. **Raise the wire bound.** Increase every owning schema/domain bound together. This postpones rather than eliminates finite overflow and has release/compatibility cost.

These are materially different protocol choices; this issue does not pre-select one.

## Acceptance criteria

- [ ] Record the chosen overflow/admission invariant in the owning ADR or `docs/INTERFACES.md` before implementation.
- [ ] Any sequence of individually accepted finding coverages either remains receipt-representable or is rejected atomically before durable history becomes unreceiptable.
- [ ] No solution silently drops distinct material gap identities without an explicit canonical overflow disclosure.
- [ ] `weakest`, `ReceiptDocument`, `receipt_weakest_coverage`, receipt construction, codecs, and JSON schemas agree on the same rule.
- [ ] Receipt failures caused by internal aggregation are never misclassified as caller `INVALID_REQUEST`.
- [ ] Focused tests cover exactly 64 gaps, 65 distinct gaps split across otherwise-valid retained findings, duplicates across findings, deterministic ordering, codec round trips, and memory/SQLite parity.
- [ ] Relevant schema fixtures, public-boundary/resource-manifest gates, Ruff, and pinned Pyright pass.

## Design gate

**Yes.** This changes protocol coverage semantics and/or durable admission behavior. Per `AGENTS.md`, wait for maintainer acknowledgement on this issue before opening an implementation PR.

## Non-goals

- Weakening the #259 requirement to retain historical finding coverage.
- Silently truncating or lexically selecting 64 gap codes.
- Treating a durable valid-state capacity failure as a malformed receipt request.

danielmkuptzov:
Fixes #262.

## What changed

- enforce the task-global 64-gap receipt coverage capacity during append admission
- allow exactly 64 distinct gaps and atomically reject the 65th with a nonretryable `LIMIT_EXCEEDED` finding
- share receipt-capacity logic across append admission, deterministic checks, and receipt generation
- preserve memory/SQLite adapter parity through the SQLite-backed memory oracle
- map legacy unrepresentable coverage to `LIMIT_EXCEEDED` instead of `INVALID_REQUEST`
- document the public capacity and error contract

## Root cause

The system bounded individual coverage declarations but did not bound their task-wide union before committing an append. A valid sequence of individually representable writes could therefore create task state whose receipt coverage could no longer be represented honestly.

## Impact

Receipt state remains representable after every accepted append. A capacity-exceeding append fails atomically and retryably safe metadata does not expose user-controlled content.

## Validation

- `uv run pytest tests/conformance/adapters/test_memory_sqlite_parity.py tests/integration/application/test_respond_status_receipt.py` — 24 passed post-commit
- broader affected receipt/check/adapter test slices — 351 passed
- `uv run ruff check .`
- `uv run ruff format --check .`
- `npx --no-install pyright` — 0 errors
- `uv run python scripts/scan_public_boundary.py --source-tree .` — 932 files passed
- `git diff --check`

> [!NOTE]
> ### Enforce a 64-code capacity limit on receipt coverage `known_gaps`
> - Adds `validate_receipt_coverage_capacity` in [`receipt_capacity.py`](https://github.com/TheGaySupreme123/yoetz/pull/263/files#diff-47f9068a8238361526ac11cbaa466057388ed9a652af6e3de9fffaf81000f86c) that computes the full union of gap codes a receipt would require and raises `ReceiptCoverageCapacityExceeded` when it exceeds `MAX_KNOWN_GAPS` (64).
> - `MemoryLedgerAdapter.append_batch` now calls this validator before committing; batches that would push `known_gaps` past 64 distinct codes are atomically rejected with `LIMIT_EXCEEDED` and `component=receipt_coverage`.
> - Duplicate gap codes do not consume capacity; rejection never truncates or reorders existing gaps.
> - `execute_receipt` now returns `LIMIT_EXCEEDED` (instead of `INVALID_REQUEST`) for legacy tasks whose retained gaps already overflow capacity.
> - Parity tests added to confirm identical behavior between memory and SQLite backends at the 64-code boundary.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3220da6.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added receipt-coverage capacity validation with a limit of 64 distinct known gaps.
  * Duplicate gaps no longer consume capacity.
  * Coverage and receipt status now consistently account for deterministic, check, storage, and finding gaps.

* **Bug Fixes**
  * Batches exceeding capacity are rejected atomically with a structured, non-retryable `LIMIT_EXCEEDED` error.
  * Receipt coverage behavior is now consistent across memory and SQLite backends.


TheGaySupreme123:
@coderabbit reivew

TheGaySupreme123 (approved):

PR #277, opened 2026-08-15
Fix Codex async hook compatibility

### LINKED ISSUE #271: [bug]: `async: true` observe hooks are dropped, not downgraded, on every stable Codex — six lifecycle events go unobserved
## Duplicate search

- [x] Searched open and closed issues and PRs for `async`, `hook`, `PreToolUse`, `skipping`, `codex version`. Closest neighbours are all **latency** issues: #209/#210/#211 (hook timeout loop, fixed by PR #213) and #242 (standing synchronous per-call cost, fixed by PR #245's advice gating + import diet). None of them covers the host **refusing to register** the handler. No duplicate.

## Area

`src/yoetz/adapters/integrations/codex_plugin.py:100-127` (`_hooks_json`, execution-mode split), `:110-113` (the incorrect host assumption), `:183-253` (six `run_async=True` sites); `tests/unit/adapters/test_codex_plugin.py:200,231,235`; `src/yoetz/cli/observe_hooks.py:1097-1105` (pre/post pairing); `src/yoetz/adapters/integrations/codex_marketplace.py:178` (the `codex --version` probe that already exists).

## Observed — Codex 0.147.0 (latest stable) and 0.146.0

Every session start prints six warnings and then silently runs without those hooks:

```
skipping async hook in ~/.codex/plugins/cache/yoetz/yoetz/0.1.0/hooks/hooks.json: async hooks are not supported yet
```

One per `"async": true` handler. Codex does not downgrade them to synchronous — it **drops them from the registry**. From `codex-rs/hooks/src/engine/discovery.rs` at tag `rust-v0.147.0` (lines 476-484):

```rust
if r#async && event_name != HookEventName::SessionEnd {
    warnings.push(format!("skipping async hook in {}: async hooks are not supported yet", ...));
    continue;                       // handler never enters the registry
}
```

`SessionEnd` is the sole exception (warn + run synchronously), which is the behaviour our comment generalises from.

On `main` the same code path reads:

```rust
let runs_async = r#async && event_name != HookEventName::SessionEnd;
```

Async command hooks landed upstream in commit `6f647caa` — "Support asynchronous command hooks (openai/codex#37533)" — committed 2026-08-08 04:24Z. `rust-v0.147.0` was tagged 2026-08-07 01:41Z, so it missed the release by ~27 hours. The first release carrying it is `0.148.0-alpha.6` (2026-08-10); current alpha is `0.148.0-alpha.19`. **No stable Codex release honours `async` today.**

### What this costs us

Dropped on every stable host: `PreToolUse`, `PermissionRequest`, `PreCompact`, `PostCompact`, `SubagentStart`, `SubagentStop`.

Still running: both `SessionStart` entries, both `PostToolUse` entries, `Stop`, `SessionEnd`, `UserPromptSubmit`.

Live corroboration from a consented workspace running 0.147.0 — `yoetz observe status --workspace .`:

```
lifecycle: stale
undelivered: 99 (cause: mapping_missing=99; last successful drain: never)
quarantine: 239
hook_diagnostics: {"count":486,...,"timings":{"count":212,"last_ms":686,"max_ms":3388}}
gaps: content_capture_unavailable,outbox_quarantined,unpaired_event
hook_coverage: True
```

`unpaired_event` is raised at `observe_hooks.py:1101-1105` when a *post* event arrives with no open *pre* for its correlation id — precisely the signature of `PreToolUse` never firing while `PostToolUse` still does. Meanwhile `hook_coverage` still reports `True`: nothing in yoetz notices that six of its eleven handlers were refused.

## Root cause

`codex_plugin.py:110-113` states:

> Codex ignores unknown handler fields, so older hosts that predate `"async"` parse this unchanged and simply run the handler sync.

That is false for every host we actually target. 0.146 and 0.147 both **recognise** the field — which is why they can emit a field-specific warning — and skip the handler. Only a host predating the field entirely would ignore it, and no such version is in use.

`tests/unit/adapters/test_codex_plugin.py:200` carries the same belief in its docstring ("Codex ignores...") and the test passes regardless, because it only asserts the JSON we emit, never what a host does with it.

## Expected

1. **No observation loss on stable hosts.** Whatever the mechanism, the six lifecycle events must reach ingest on Codex ≤ 0.147.
2. **Detect the refusal.** If the host will not run a declared handler, yoetz should say so once and `hook_coverage` should not report `True`. (#242 asked for the weaker version of this — detecting a sync *downgrade* — and it was never built.)
3. **Correct the comment and the test rationale** so the assumption is not re-derived later.

## Options considered

**A. Drop `async` entirely (sync everywhere).** Correct on every host, no gating. Measured cost of the loaded pass on the live store above is ~0.7 s typical / 3.4 s worst per hook, ×2 hooks per tool call — a partial regression toward #242.

**B. Host-capability gate at activation.** Emit `"async": true` only for hosts ≥ 0.148. The probe already exists (`codex_marketplace.py:178` captures `codex_version` into the preview/plan before the tree is rendered at `:755-774`). Costs: `render_plugin_tree`/`render_plugin_install_tree` gain a capability parameter that must be threaded through `_source_cache_members` (`:447`), `_plugin_source_digest` (`:737`) **and every `_assert_plugin_source` re-render** (`:955`, called at `:1052-1090`) or activation fails with `PREVIEW_STALE`; `plugin_source_digest` becomes host-dependent (surfaced in `setup.py:719`, `tui/render.py:273,309`, recorded at `setup.py:1362,1374`); a Codex upgrade across the 0.148 line makes the installed tree read `INSTALLED_UNTRUSTED_UNKNOWN` → `IntegrationState.MODIFIED` (`codex_plugin.py:56-80`) and re-triggers Codex hook-trust approval; and `tests/packaging/test_committed_plugin_tree.py` needs an explicit decision about which variant is the committed one. Note `profile_for_codex_version` (`adapters/importers/codex_jsonl.py:370`) cannot be reused — it is an exact-match table that refuses to infer ranges — so the gate needs its own comparator with prerelease handling and an unknown-version-means-legacy default.

**C. Detached handoff in the hook itself (preferred).** All six handlers are pure ingress that always emit `{}`. The hook writes its envelope, hands the remaining work to the daemon, and exits — non-blocking on *every* Codex version, with no capability gate and no host-dependent plugin tree. Gotcha: the detach must not leave stdio open (double-fork + devnull), or the host waits on the pipe regardless of how fast the parent exits. This is what #242's "Expected" section already asked for ("persistent lightweight forwarder / daemon-side tap"), and it demotes `async` from a correctness dependency to an optimisation.

Measurements supporting C: the hook entry path is no longer startup-dominated after PR #245's import diet — an isolated pass that bails at the consent gate costs **0.05 s** (for reference `yoetz --version` costs 0.28 s). The remaining ~0.7 s is work *inside* the pass (ingest, the 0.20 s drain, advice refresh), which is exactly what a daemon handoff removes.

## Severity

**High.** Silent observation loss on 100% of stable Codex installs, for the events that carry tool-call and compaction coverage, with the health surface still reporting `hook_coverage: True`.

danielmkuptzov:
Fixes #271

## What changed

- Detect async command-hook support from the exact probed Codex version.
- Emit `"async": true` only for Codex `>= 0.148.0-alpha.6`.
- Fail closed to bounded synchronous handlers for stable legacy, missing, malformed, and oversized versions so required lifecycle events are slower rather than silently dropped.
- Thread the selected version through managed source rendering, ownership markers, activation previews, source/cache digests, setup, TUI presentation, and every apply-time integrity fence.
- Allow a byte-exact managed variant transition to be previewed, while refusing activation before mutation until the intended source variant is installed.
- Regenerate the committed plugin tree through its owning installer; the unprobed committed variant is conservatively synchronous.
- Document the behavior and trust-boundary implications in ADR-012 and `docs/INTERFACES.md`.

## Root cause

Stable Codex recognizes the async handler field but skips non-`SessionEnd` command handlers that declare it. Yoetz assumed older hosts ignored the field and ran those handlers synchronously, causing six observation handlers to disappear from the registry.

## Impact

Codex versions without async command-hook support now retain `PreToolUse`, `PermissionRequest`, `PreCompact`, `PostCompact`, `SubagentStart`, and `SubagentStop` through the synchronous fallback. Capable hosts retain the lower-latency async path.

Version-dependent bytes remain inside the existing preview, source-integrity, cache-integrity, and explicit activation-approval boundaries. Unknown versions never opt into the unsafe optimization.

## Verification

- `uv run pytest tests/packaging/test_committed_plugin_tree.py tests/unit/adapters/test_codex_plugin.py tests/unit/adapters/test_codex_marketplace.py tests/unit/cli/test_setup_activation.py tests/unit/tui/test_runtime.py -q` — 72 passed
- Broader observation slice — 168 passed; one unrelated existing failure: `test_session_start_creates_binding_without_mcp_start` expects advisory text but receives `{}`, reproduced alone
- `npx --no-install pyright` — 0 errors
- Ruff lint and format checks — passed
- `git diff --check` — passed
- `uv run --locked python scripts/scan_public_boundary.py --source-tree .` — 940 files scanned, passed

## Process

- Duplicate issue/PR search completed; no matching open PR found.
- Packaging behavior is documented in the amended ADR and shared interface documentation.
- The committed generated plugin files were regenerated through `install_plugin`, not hand-edited.

> [!NOTE]
> ### Fix Codex async hook compatibility by gating async hooks on Codex version
> - Introduces `codex_supports_async_hooks` in [`codex_plugin.py`](https://github.com/TheGaySupreme123/yoetz/pull/277/files#diff-851d764d011f060fa95dbe21ed22cccdfc954ecdaad2aea017cbb03a2a66a933) to check whether the probed Codex binary is ≥ `0.148.0-alpha.6`.
> - `_hooks_json` now conditionally emits `"async": true` on pure-ingress observe handlers only when the version check passes; older versions get synchronous handlers with a 10-second timeout.
> - Plugin tree rendering, inspection, and installation throughout the stack now accept and forward `codex_version` so the correct variant is written and validated.
> - Variant transitions (sync ↔ async) can be previewed but `apply` is fenced until the installed source tree matches the intended variant.
> - Behavioral Change: installations against older Codex versions will no longer write async hook declarations, and `inspect`/`doctor` results now reflect the version-specific expected tree.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9657650.</sup>
> 
>


TheGaySupreme123:
@macroscope full code review

PR #281, opened 2026-08-15
Read the observation runtime gate lock-free; never silently discard a hook event at the gate

### LINKED ISSUE #273: [bug]: store-lock contention is reported as runtime_gate_unsafe and silently discards the whole hook event — currently 100% of events on a consented workspace
## Duplicate search

- [x] Searched open and closed issues/PRs, and full-text scanned every issue body, for `runtime_gate_unsafe`, `store lock`, `lock_timeout`, `contention`, `runtime gate`. Zero hits. #209/#210/#211/#213 are hook *latency*; #223/#232 are service-side livelock/saturation; #204 is a delivery outage with the hook still ingesting. None covers the hook refusing to capture at all. No duplicate.

## Area

`src/yoetz/cli/observe_hooks.py:1029-1034` (the runtime-gate guard in `handle_observe`), `src/yoetz/adapters/integrations/observation_local.py:705-753` (`runtime_enabled`, which takes the store lock at `:709`), `:109` and `:152-177` (`_STORE_LOCK_TIMEOUT_SECONDS = 2.0`, `_InterprocessStoreLock.__enter__`).

## Observed — live diagnostics, this machine, 2026-08-15, `main` (`86a776e`)

`~/Library/Application Support/yoetz/observation/hook-diagnostics.jsonl`, current ring (197 records, 2026-08-14T19:32Z → 2026-08-15T18:25Z):

| reason | count |
| --- | --- |
| `runtime_gate_unsafe` | **45** |
| `hook_budget_exceeded` | 37 |
| `drain_lease_contended` | 28 |
| `drain_budget_exhausted` | 16 |
| *(others)* | — |

`runtime_gate_unsafe` is spread across `PreToolUse` (19), `PostToolUse` (25) and `PostCompact` (1), and it is **the last 8 consecutive records** (18:23:00Z → 18:25:41Z) — i.e. at the time of writing every hook event on this machine is being discarded.

Meanwhile the gate file is entirely healthy:

```
$ ls -l .../observation/runtime-gate.json
-rw-------  1 shayb  staff  61 …
$ cat .../observation/runtime-gate.json
{"enabled":true,"schema":"yoetz.observation-runtime-gate/1"}
```

and a direct call returns `True`. Nothing is unsafe. The name of the diagnostic is wrong, and so is the action it triggers.

## Root cause

`runtime_enabled` reads the gate **inside the interprocess store lock** (`observation_local.py:709`). `_InterprocessStoreLock.__enter__` raises `TimeoutError("observation_store_lock_timeout")` after `_STORE_LOCK_TIMEOUT_SECONDS = 2.0` when another process holds it. `handle_observe` catches *every* exception from that call and treats it as an unsafe gate:

```python
try:
    capture_enabled = store.runtime_enabled()
except Exception:
    record_hook_diagnostic("runtime_gate_unsafe", resolved_event, _state=_state)
    _stdout_json({}, stdout)
    return 0
```

That `return 0` is before session binding, ordinal allocation, local ingest, outbox enqueue, and stream reconcile. The event is not deferred and not spooled — it is **gone**, with no `hook_observe_degraded:` stderr line (unlike every other degraded branch in this function) and no workspace gap code. The only trace is an opaque counter inside the `hook_diagnostics` blob in `observe status`.

Holders routinely exceed 2s: the whole local pass runs inside `store.batched(workspace_commitment)` (`observe_hooks.py:1108`), and #271's own live timings report `max_ms: 3388`. The store's own docstring notes Codex runs up to 8 hooks concurrently (`observation_local.py:1421-1429`), so the more the async work in #271 succeeds, the more contention this guard converts into data loss.

Deterministic repro (isolated state root, no real store touched):

```python
s = LocalObservationStore(_state=tmp)
s.set_runtime_enabled(True)
# child process: `with s._lock: sleep(3)`
s.runtime_enabled()
# -> TimeoutError: observation_store_lock_timeout, after exactly 2.00s
```

## Expected

- Lock contention is not "unsafe". Distinguish `TimeoutError` from `PublicOperationError(STORAGE_UNSAFE)` and give them different diagnostics; only the latter should fail closed.
- Read the gate without serializing on the workspace store lock (it is a 61-byte owner-only file with its own validation), or retry the acquisition for the hook's remaining budget.
- If an event is dropped for any reason, it is recorded as a workspace gap so `observe status` and coverage wording can see it — a silently missing event is the one outcome this subsystem must not produce.

## Severity

**High.** Total, silent observation loss under ordinary concurrency, currently affecting 100% of hook events on a consented dogfood workspace while every health surface reports the workspace as consented and covered.

Found while investigating the live-evidence block in #271; it is independent of the async-hook drop described there and is made worse by fixing it.

TheGaySupreme123:
Fixes #273.

## Root cause

`runtime_enabled` read the 61-byte gate marker inside the interprocess store lock. The lock's 2-second acquisition timeout is routinely exceeded by a concurrent hook's `store.batched(...)` local pass (Codex runs up to 8 hooks concurrently; #271 measured `max_ms: 3388`), and the guard in `handle_observe` caught the resulting `TimeoutError` under `except Exception`, labeled it `runtime_gate_unsafe`, and returned before session binding, ingest, and outbox enqueue — silently discarding the event with no stderr line and no workspace gap.

## Changes

- **`observation_local.runtime_enabled`** no longer takes the store lock. The marker is an owner-only file that `set_runtime_enabled` only ever replaces atomically, so the lock bought nothing but contention. The read now goes through a single descriptor (`O_NOFOLLOW` open + `fstat` + bounded read), which also closes the old lstat-then-read TOCTOU: the safety checks and the payload are bound to one inode. Validation semantics are unchanged — missing marker defaults to enabled; symlink, foreign owner, loose mode, bad size, or bad shape still raise `STORAGE_UNSAFE` and fail closed.
- **`handle_observe`** distinguishes the two failures the issue conflated:
  - `TimeoutError` (store-lock contention, now defense-in-depth) records `runtime_gate_contended` plus a `hook_observe_degraded:` stderr line and falls back to the missing-marker default (enabled) instead of discarding the event. Consent still gates every ingest downstream.
  - Any other failure still fails closed as `runtime_gate_unsafe`, but now emits the `hook_observe_degraded:` stderr line like every other degraded branch and best-effort records an `observation_storage_corrupt` coverage gap for the bound workspace (explicit workspace argument first, then the Codex-session map, consent-checked), so `observe status` and coverage wording can see the drop. No new wire gap code was introduced.
- **`hook_diagnostics`** admits the new `runtime_gate_contended` reason.

## Verification

- New regression test holds the store lock from another thread and asserts `runtime_enabled()` returns immediately; the issue's cross-process repro (child process holding the flock for 5s) now answers in <1ms where it previously raised `TimeoutError` after exactly 2.00s.
- New tests: contention at the gate still ingests the event (outbox pending, `runtime_gate_contended` recorded, no `runtime_gate_unsafe`); an unsafe gate on a consented workspace records the `observation_storage_corrupt` gap; the existing fail-closed test now also asserts the stderr line.
- `tests/unit/cli` (337), `tests/unit/adapters` + `tests/integration/service/test_ready_composition.py` (269), and `tests/integration/observation` all pass; the one failure there (`test_session_start_creates_binding_without_mcp_start`) fails identically on a clean `main` checkout on this machine and is unrelated. Ruff and pinned Pyright are clean on the touched files.

> [!NOTE]
> ### Read the runtime gate lock-free and never silently discard hook events on gate contention
> - [`LocalObservationStore.runtime_enabled`](https://github.com/TheGaySupreme123/yoetz/pull/281/files#diff-dcc1feb8bc7f834d09056738e52fec58f85bdd06640418a241feda997cb5be82) is rewritten to open the gate file with `O_RDONLY | O_CLOEXEC | O_NOFOLLOW` (non-blocking), validate ownership/permissions via `fstat`, and read with a hard size cap — all without acquiring the interprocess store lock.
> - On store-lock contention (`TimeoutError`), `handle_observe` in [`observe_hooks.py`](https://github.com/TheGaySupreme123/yoetz/pull/281/files#diff-cdfa4cdbe7cfccf96a65db7f69d192cac91fad8eba4ec777e16f9a4d177f49c4) now logs `runtime_gate_contended`, records a diagnostic, and proceeds as if capture is enabled instead of dropping the event.
> - Unsafe gate conditions (symlinks via `ELOOP`, oversized content, bad ownership) emit a stderr line, record a `runtime_gate_unsafe` diagnostic, and call the new `_note_dropped_event_gap` helper to record an `OBSERVATION_STORAGE_CORRUPT` coverage gap for the bound workspace.
> - Event correlation in `handle_observe` now prefers `tool_use_id`, then `tool_call_id`, then `correlation_id` for pre/post event pairing.
> - Behavioral Change: gate evaluation no longer blocks on the interprocess lock; FIFO or symlink gate paths now fail closed rather than blocking or following the link.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6ce4504.</sup>
> 
>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -550,6 +550,16 @@ carried them; they are listed because each one describes the behavior that now s
   session where every check returned `blocked_by_policy` / `route_semantic_ceiling` with nothing
   connecting the two. `yoetz privacy setup` now names the mismatch and the command that fixes it,
   and the terminal interface reports the agent-route verdict as its own readiness line.
+- Ordinary store-lock contention discarded hook events as `runtime_gate_unsafe`: the runtime-gate
+  read serialized on the interprocess store lock, whose two-second acquisition timeout is routinely
+  exceeded by a concurrent hook's batched local pass, and the guard treated the resulting
+  `TimeoutError` as an unsafe gate — dropping the event before capture with no stderr line and no
+  workspace gap, on a workspace every health surface reported as covered. The gate is now read
+  lock-free through a single descriptor (it is an owner-only marker replaced only atomically), a
+  contended read is reported as `runtime_gate_contended` and falls back to the missing-marker
+  default instead of discarding the event, and a genuinely unsafe gate still fails closed but says
+  so on stderr and records an `observation_storage_corrupt` coverage gap for the bound workspace so
+  `observe status` can see the drop (issue #273).
 
 ### Security
 
--- a/src/yoetz/adapters/integrations/observation_local.py
+++ b/src/yoetz/adapters/integrations/observation_local.py
@@ -715,7 +715,7 @@ class LocalObservationStore:
         """
 
         path = self._root / _RUNTIME_GATE_NAME
-        flags = os.O_RDONLY | os.O_CLOEXEC
+        flags = os.O_RDONLY | os.O_CLOEXEC | getattr(os, "O_NONBLOCK", 0)
         if hasattr(os, "O_NOFOLLOW"):
             flags |= os.O_NOFOLLOW
         try:
@@ -749,6 +749,12 @@ class LocalObservationStore:
                 chunks.append(chunk)
                 remaining -= len(chunk)
             raw = b"".join(chunks)
+            if len(raw) > _MAX_RUNTIME_GATE_BYTES:
+                raise _error(
+                    PublicErrorCode.STORAGE_UNSAFE,
+                    "Observation runtime gate is unsafe.",
+                    retryable=False,
+                )
         except OSError as exc:
             raise _error(
                 PublicErrorCode.STORAGE_UNSAFE,
--- a/src/yoetz/adapters/integrations/observation_local.py
+++ b/src/yoetz/adapters/integrations/observation_local.py
@@ -13,6 +13,7 @@ import dataclasses
 import errno
 import os
 import re
+import stat
 import threading
 import time
 from collections.abc import Callable, Generator, Mapping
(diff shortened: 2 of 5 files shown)


TheGaySupreme123:
@macroscope full code review

PR #283, opened 2026-08-15
Fix observation drain reordering, immortal unmapped-session rows, and terminal soft-lock reverts

### LINKED ISSUE #272: [bug]: observation drain reorders its own queue — every recovered session loses SessionStart, the first prompt, and its first tool pair to a terminal cursor_stale quarantine
## Duplicate search

- [x] Searched open and closed issues/PRs for `quarantine`, `cursor_stale`, `outbox`, `ordering`, `sweep`, `attempts`. The neighbours are all adjacent, none covering: #211 bounded quarantine *growth* (retention/eviction), #218 drain *throughput*, #219 latched *gap reporting* (it observed `cursor_stale` in `gaps` and treated it as a stale label, not as live loss), #204 delivery outage. Full-text scan of every issue body for `cursor_stale` returns #210, #219, #259, #262 — none of them names the delivery-ordering cause or the terminal classification. No duplicate.

## Area

`src/yoetz/application/observation_drain.py:276-291` (`ObservationOutboxSweeper._fair_pending_rows`, the intra-lane sort at `:284`), `:41-49` and `:82-86` (`RETRYABLE_OBSERVATION_REJECTIONS` / `route_observation_ingest`), `src/yoetz/adapters/sqlite/observation.py:145-160` (monotonic cursor rejection) and `:919-945` (`_note_gap_event`), `src/yoetz/cli/observe_hooks.py:868-876` (the hook drain's same-session continue after `service_unavailable`).

## Observed — live store, read-only inspection 2026-08-15, `main` (`86a776e`)

Workspace `/Users/shayb/yoetz-core`, commitment `hmac-sha256:722a4470…2540`.

`quarantine` holds 243 entries. Excluding 196 legacy `setup_probe` rows, **47 carry reason `cursor_stale`**, dated 2026-08-12 → 2026-08-15, and every one is an *opening* event of a real session:

| event_kind | quarantined |
| --- | --- |
| PreToolUse | 19 |
| PostToolUse | 14 |
| SessionStart | 7 |
| UserPromptSubmit | 7 |

Worked example — Codex session `01a006a4-2ac1…` (session commitment `hmac-sha256:cce1e10e…`, mapped to task `tsk_c07dcb6a…`):

- Local quarantine holds its `event_position` 1–4 envelopes (SessionStart 18:18:42Z, UserPromptSubmit 18:18:44Z, PreToolUse 18:18:53Z, PostToolUse 18:18:54Z), reason `cursor_stale`, `quarantined_at` 18:20:01–18:20:03Z.
- The task bundle's `observation_events` table starts at **`event_position` 5** (rowids 1–9 are positions 5…13, 18:19:01Z → 18:19:50Z). Positions 1–4 appear only as rowids 10–13, `event_kind = "observation_gap"`.

So the session's SessionStart, its first user prompt, and its first tool-call pair were **destroyed and replaced by gap placeholders**, while everything after them was ingested normally. This is not a hook that failed to fire — the envelopes exist, were captured correctly, and were then thrown away by our own delivery path.

## Root cause

1. While the session had no ledger mapping, its rows were rejected `mapping_missing` (`application/observation_coordinator.py:456`) and their `attempts` counters climbed (`observation_local.py:1346-1377`).
2. `_fair_pending_rows` orders each session lane by `(attempts, index)`:

   ```python
   lanes[(workspace, session)] = [
       row for _, row in sorted(rows, key=lambda item: (item[1].attempts, item[0]))
   ]
   ```

   and then picks lanes by `lanes[item][0].attempts` (`:288-297`). Fairness *across* lanes is the intent, but the same key is applied *within* a lane, so a newer, less-attempted row of one session is delivered ahead of an older, more-attempted row of that same session. Once the mapping appeared, position 5 (attempts 0) went first.
3. Ingest enforces a monotonic per-`(workspace, source, session)` cursor and rejects anything behind it as `cursor_stale` (`adapters/sqlite/observation.py:145-160`). Correct in isolation — it is the delivery order that was violated.
4. `cursor_stale` is **not** in `RETRYABLE_OBSERVATION_REJECTIONS`, so `route_observation_ingest` returns `QUARANTINE` (`observation_drain.py:82-86`). Quarantine is terminal: the envelope leaves the delivery path permanently.
5. `_note_gap_event` writes an `observation_gap` row at that cursor, so the ledger reads as "observed, with a gap" rather than "we reordered our own queue and dropped four events".

The hook-side drain preserves intra-session FIFO (`observe_hooks.py:771-790`), so the reordering originates in the service sweeper; the quarantining is then executed by whichever drain next reaches the stranded rows. The hook drain has an independent door to the same outcome: after a single `service_unavailable` on row *N* it continues to row *N+1* of the **same** session (`observe_hooks.py:868-876`) instead of retiring that session's lane for the pass, which likewise puts a later event in front of an earlier one.

## Expected

- Per-session delivery is FIFO in **every** drain path. Fairness may reorder lanes; it must never reorder within a lane. A lane whose head row fails is skipped for the pass, not stepped over.
- A `cursor_stale` rejection for a row that has **never** been acknowledged is treated as a delivery-order defect (retry/repair, or an explicit loud terminal outcome), not silently absorbed into quarantine alongside genuinely invalid envelopes.
- If an event is going to be replaced by an `observation_gap` placeholder, that substitution is visible in `yoetz observe status`, not just as a quarantine depth number.

## Severity

**High.** Silent, permanent loss of exactly the events that anchor a session — `SessionStart`, the first `UserPromptSubmit`, and the first tool-call pair — in the subsystem whose purpose is a trustworthy record. 47 confirmed lost on one machine. The loss is deterministic for any session that starts before its ledger mapping exists, which is the common case (see the companion issue on unmapped-session rows).

Found while investigating the live-evidence block in #271; it is not caused by the async-hook drop described there.

### LINKED ISSUE #275: [bug]: outbox rows from sessions that ended unmapped are immortal — 99 rows, 194 attempts each, 6 days old, holding 19% of the outbox bound
## Duplicate search

- [x] Searched open and closed issues/PRs for `mapping_missing`, `outbox`, `backlog`, `attach`, `auto-start`, `retry`, and full-text scanned every issue body. #219 fixed *reporting* a `mapping_missing` gap after the mapping exists — not the rows. #218 is throughput for rows that are deliverable. #211 bounded quarantine, not the pending outbox. #204 was a total delivery outage, closed by #208. None gives an unmapped session's rows a terminal state. No duplicate.

## Area

`src/yoetz/cli/observe_hooks.py:1213-1240` (auto-attach is attempted only on `SessionStart`), `:897-965` (`_try_auto_start`); `src/yoetz/application/observation_coordinator.py:454-457` (`mapping_missing` rejection); `src/yoetz/application/observation_drain.py:41-49` (`RETRYABLE_OBSERVATION_REJECTIONS`) and `:230-240` (only `observation_storage_corrupt` retires a session); `src/yoetz/adapters/integrations/observation_local.py:84` (`_MAX_OUTBOX = 512`) and `:1285-1288` (overflow drops new work).

## Observed — live store, read-only inspection 2026-08-15, `main` (`86a776e`)

Workspace `/Users/shayb/yoetz-core`, commitment `hmac-sha256:722a4470…2540`:

- `pending_outbox`: **99 rows, every single one `last_reason = "mapping_missing"`**
- every row's `attempts` counter reads **194**
- receipt times span **2026-08-09 → 2026-08-13** — six days old at the time of writing
- spread over 6 Codex sessions; 8 of the rows are `SessionEnd` events, so those sessions demonstrably ended long ago
- kinds: `UserPromptSubmit` 47, `PreToolUse` 20, `PostToolUse` 12, `SessionEnd` 8, `SessionStart` 6, `Stop` 6

These rows can never be delivered. The mapping is keyed by Codex session id; those sessions are over and no future `start` will ever produce one. Yet they are re-attempted on every sweep, forever, and they hold 99 of the 512 outbox slots permanently.

## Root cause

1. **Attach is tried once, at `SessionStart` only.** `handle_observe` calls `_try_auto_start` inside the `SessionStart` branch under a session lock (`observe_hooks.py:1226-1234`). Every later event only does `load_mapping`. If the service is not reachable at session start — the common case, since the daemon spawns on demand — `_try_auto_start` returns `None` (it swallows every exception, `:963-964`) and the session is unmapped **for its entire life**. The hook does emit "Call start to attach a task" as additional context, but nothing re-attempts it.
2. **`mapping_missing` is classified retryable forever.** It is in `RETRYABLE_OBSERVATION_REJECTIONS` (`observation_drain.py:41-49`), so `route_observation_ingest` always returns `RETRY`. There is no attempt cap, no age bound, and no terminal disposition. The hook drain at least retires the session's lane for the current pass (`observe_hooks.py:878-890`); the service sweeper has no such rule — it retires a session only on `observation_storage_corrupt` (`observation_drain.py:230-240`) — which is how a single row reaches 194 recorded attempts.
3. **The store knows the session ended and does not use it.** `note_session_end` records ended sessions (`observe_hooks.py:1163-1169`) and `_status_unlocked` reads them for lifecycle, but nothing consults them when deciding whether an outbox row still has a future.

Downstream effects, all currently visible on this machine:

- `yoetz observe status` reports `undelivered: 99 (cause: mapping_missing=99)` indefinitely, so the number can no longer be read as "work in flight".
- 99/512 of the outbox bound is permanently consumed by dead rows. At the bound, `enqueue_outbox` returns `outbox_overflow` and **new** observations are dropped at ingress (`observation_local.py:1285-1288`) — dead rows can eventually disable observation for a live workspace.
- Every sweep pays a service round trip per dead row, and every `_load`/`_save` re-encodes them — the recurrence tax pattern #211 named, in a different bucket.

## Expected

- An unmapped session that has ended is terminal: retire its rows with an explicit reason (quarantine with `mapping_missing`, or expire by age) so the pending outbox has a bounded terminal state and `undelivered` means what it says.
- Re-attempt attach on later events, not only `SessionStart`, so a service that was down for the first 200ms of a session does not cost that session's entire record.
- Bound retries by attempts or age regardless, so no row can accrue 194 identical failures.

## Severity

**Medium-High.** Not itself data loss today, but it permanently degrades the health surface, wastes a fixed capacity budget that has a hard failure mode at the bound, and it is the precondition for the delivery-ordering loss filed alongside this issue — the early events destroyed there are exactly the rows that were stranded here first.

Found while investigating the live-evidence block in #271; it is independent of the async-hook drop described there.

### LINKED ISSUE #276: [bug]: a succeeded soft-lock re-ready is reverted by a transient activation failure, and VAULT_LOCKED misreports it as a non-retryable hard lock
## Duplicate search

Searched open and closed issues for `vault`, `unlock`, `relock`, `idle`, `lock`, `receipt`. Nearest neighbours: #21/#23/#26 (auto-unlock provisioning and stale-passphrase repair — all about *provisioning* the secret, not about a successful unlock being reverted), #157/#111 (retained passphrase vault survival and wizard initialization), #273 (store-lock contention discarding hook events — a different lock, on the observation store, not the vault), and the receipt-blocking family #216/#231/#259/#262 (all reject at receipt-build time, none prevent the ledger write itself). None covers a *succeeded* soft-lock re-ready being rolled back by a later failed activation. No duplicate.

## Area

`src/yoetz/service/daemon.py:1613-1630` (the `except Exception:` that logs `unlock_failed`, force-locks the vault, and returns to LOCKED after `vault.unlock()` has already succeeded); `:1700-1711` (the `soft_lock_reready` success path); `src/yoetz/service/loop_health.py:106-121` (`control_plane_saturation_entered` / `_cleared`); the `VAULT_LOCKED` public error surfaced by `mcp.bridge`.

## Observed — live dogfood, 2026-08-15

A Codex session (`01a006a4-2ac1-7050-835a-897e0ef5a955`) opened a task, published its plan and three obligations successfully at 18:21:56Z, then implemented and verified for ~22 minutes. At 18:43:59Z it issued its first post-implementation Yoetz call to record evidence and was refused with `VAULT_LOCKED`. It abandoned evidence publication, `check`, and `receipt`, and reported to the user that no completion receipt was available.

The service log shows the vault was **not** hard-locked. From `~/Library/Logs/yoetz/service.stderr.jsonl` and `service.diagnostics.jsonl`:

```
18:43:57.313Z service.daemon auto_unlock outcome=succeeded reason=soft_lock_reready
18:43:59.028Z service.daemon control_plane_saturation_entered reason=event_loop_lag duration_ms=3996 operation_count=1
18:43:59.396Z service.daemon auto_unlock outcome=failed  reason=unlock_failed
18:43:59.439Z mcp.bridge     status_public_error         reason=vault_locked  request_id=req_a2f70647-...
18:44:00.032Z service.daemon control_plane_saturation_cleared reason=event_loop_lag duration_ms=1004 operation_count=0
```

The soft lock re-readied successfully at 18:43:57. Two seconds later, under a ~4 s event-loop lag, a subsequent soft-unlock's ready-application activation threw; the handler at `daemon.py:1613` caught it, logged `unlock_failed`, force-locked the vault, and returned the service to LOCKED. The saturation **cleared 600 ms after** the agent received the error.

`yoetz service status` immediately afterwards:

```json
{"state":"locked","state_reason":"unlock_failed","vault_mode":"passphrase",
 "idle_relock_seconds":900,"session_monitor":"unavailable","service_generation":"57"}
```

### The error text is wrong on both counts

The agent received:

> The local service vault is locked or uninitialized. After trusted setup with platform auto-unlock, soft locks (idle/session) re-open automatically; **this error means a hard lock or missing setup.** On a local terminal run `yoetz service unlock` … — with `"retryable": false`.

It was neither a hard lock nor missing setup: it was a soft lock that *had already successfully re-readied* and was then knocked back down by a transient 4-second event-loop lag. `retryable: false` told the agent not to try again, when retrying ~1 s later would have succeeded. The advice to run a manual terminal ceremony is also unactionable for a sandboxed agent — this one tried and was refused by its permission profile (`uv` cache: `Operation not permitted`).

### What drove the saturation

`observation_sweep_failed` fired continuously throughout the session — 18:22:07, 18:23:14, 18:25:38, 18:26:40, 18:28:00, 18:29:10, 18:33:09, 18:34:17, 18:40:44, 18:41:49 — while the workspace carried `undelivered: 99` (all `mapping_missing`, "last successful drain: never") and `quarantine: 243`. The daemon was thrashing the same undrainable backlog described in #272 and #275, alongside the hook storm in #273.

## Root cause

`daemon.py:1613-1630` treats any exception from `_activate_ready_application_locked` as an unlock failure and reverts an already-successful vault unlock, with no retry and no distinction between a transient activation failure (event-loop saturation, contention) and a genuine credential or setup failure. The resulting public error then asserts a terminal cause (`hard lock or missing setup`, `retryable: false`) that the daemon's own logs contradict.

## Expected

1. A transient ready-application activation failure should not revert a vault unlock that has already succeeded, or should retry before doing so — especially while `control_plane_saturation_entered` is active.
2. `VAULT_LOCKED` arising from `unlock_failed` under saturation should be `retryable: true` and should say so, rather than asserting a hard lock and prescribing a manual ceremony.
3. The error text should distinguish "hard lock / missing setup" from "transient activation failure"; the daemon already has `state_reason` and the saturation signal to tell them apart.
4. Failing 1-3, the loss should at minimum be visible: an agent that cannot publish evidence, `check`, or `receipt` currently discovers this only by having its call rejected after the work is finished.

## Severity

**High.** This is the integrity-critical tail of the protocol. The failure lands precisely at evidence publication, `check`, and `receipt` — after the work is complete — and converts a fully verified session (72 focused tests, Pyright clean, Ruff clean, 940-file public-boundary scan) into one with no receipt and no recorded evidence. The condition was transient and self-cleared within one second, so the entire loss was avoidable.

TheGaySupreme123:
Fixes #272, fixes #275, fixes #276.

## #272 — drain reordered its own queue, destroying session-opening events

- `_fair_pending_rows` no longer sorts within a session lane by attempts: fairness may reorder *lanes*, never rows *within* a lane. This was the root reorder that advanced the ingest cursor past a recovered session's older rows and destroyed them as terminal `cursor_stale` quarantine (47 confirmed live).
- A lane whose head row fails retryably is retired for the pass — never stepped over — in **both** drain paths: the service sweeper (all retryable reasons) and the hook drain (`service_unavailable` previously continued into the same session). Retired siblings are stamped with the shared cause so `observe status` never reports `not_attempted`.
- The sweeper re-reads lane order under the drain lease and attempts a selected row only while it is still its session's head, closing the selection-to-lease race with a concurrent hook drain (the previous `row in pending_rows` check silently skipped a bumped head and delivered its successor).
- Visibility (Expected §3): `yoetz observe status` now reports `quarantine_causes` (JSON) / `cause: cursor_stale=N,...` (text), so an event destroyed and replaced by an `observation_gap` placeholder is visible as its cause. `cursor_stale` routing stays terminal — with FIFO restored it can no longer be produced by our own reordering, and making it retryable would let genuinely stale rows retry forever — but it is now loud instead of silently absorbed.

## #275 — outbox rows of sessions that ended unmapped were immortal

- Both drain paths now treat a `mapping_missing` rejection for a session the store has marked ended as **terminal**: the whole lane is quarantined with reason `mapping_missing` (visible in the new cause breakdown, reclaimable via `observe reclaim`), freeing the outbox bound instead of holding it forever (99/512 slots, 194 attempts/row live). A restarted session generation clears the ended mark, so only genuinely dead sessions retire.
- Auto-attach is re-attempted after `SessionStart` on turn-boundary events only (`UserPromptSubmit`, `Stop`, `SessionEnd`) under the session lock with a 1.0s budget, so a daemon that missed a session's first 200ms no longer costs that session's entire record. `PreToolUse`/`PostToolUse` storms never retry.
- The blanket attempt/age cap from Expected §3 is deliberately **not** added: lane retirement (#272) already stops per-row attempt accrual, ended-session terminality removes the immortal case, and an age bound could destroy deliverable rows of a live session during a long service/vault outage. The residual is a session that crashes without its `SessionEnd` hook — happy to add an age bound in a follow-up if you want one.

## #276 — a succeeded soft-lock re-ready was reverted into a terminal lock

- `_try_soft_lock_auto_ready` now distinguishes activation failures after a **successful** vault unlock from credential failures: the soft-lock reason (and with it auto-ready eligibility) is preserved for up to 3 consecutive attempts, so the very next dispatch retries — addressing the follow-up finding that nothing ever reconsidered. A persistent activation fault still degrades to terminal `unlock_failed`; credential-shaped failures are unaffected. The counter resets on successful activation, including via the ceremony path.
- The daemon's `VAULT_LOCKED` error now carries honest retryability (true only while the lock can self-heal), and the MCP bridge stops discarding it — previously every `vault_locked` reached agents as `retryable: false` regardless of what the daemon said.
- The bridge message now branches on that flag: a retryable lock says "re-opens automatically, retry shortly" instead of asserting "hard lock or missing setup" and prescribing a manual ceremony a sandboxed agent cannot perform. The hard-lock text is unchanged for genuinely terminal locks.
- Re-evaluation is on-demand at dispatch rather than wired to the saturation-clear signal: the state stays eligible, so the next agent call (the one that would have failed) is itself the retry.

## Verification

- Full suite: 3476 passed; the only 3 failures are this machine's known pre-existing baseline (AF_UNIX endpoint tests + the SessionStart acceptance test), unrelated to this change and passing in CI.
- Pyright clean, Ruff (lint + format) clean.
- New regression tests: FIFO-despite-attempt-skew, head-retry-never-steps-over (both drains), selection-to-lease race retirement, ended-session terminal quarantine (both drains), turn-event-only attach retry, transient-activation retryability + bounded degradation, and the retryable/terminal `VAULT_LOCKED` message split.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved observation delivery to preserve per-session ordering and prevent blocked sessions from delaying unrelated activity.
  - Ended, unmapped sessions are now quarantined instead of retried indefinitely.
  - Added bounded retries for automatic session attachment and temporary service failures.
  - Improved recovery from temporary vault unlock and activation failures, with terminal failures reported clearly.
  - Vault-locked errors now provide guidance appropriate to whether retrying may succeed.

- **Improvements**
  - Observation status now shows quarantine counts by cause in JSON and human-readable output.

> [!NOTE]
> ### Fix observation drain FIFO reordering, immortal unmapped-session rows, and terminal soft-lock reverts
> - Reworks `ObservationOutboxSweeper` to preserve strict per-session FIFO order, retire session lanes on retryable failures, and stop workspace passes early on workspace-global reasons (`vault_locked`, `observation_disabled`, `paused`).
> - Adds `quarantine_ended_unmapped_session` to `LocalObservationStore` to atomically quarantine outbox rows for ended unmapped sessions, fenced by the session lifecycle lock to avoid races with concurrent attach.
> - Bounds `_try_soft_lock_auto_ready` retries to 3 consecutive activation failures before degrading to `unlock_failed`; transient failures now restore soft-lock state and remain retryable instead of becoming permanently locked.
> - `VAULT_LOCKED` responses from the MCP server and daemon `dispatch` now carry an accurate `retryable` flag reflecting whether soft auto-ready is still eligible.
> - Adds bounded auto-attach retries on specific turn-boundary hook events with `auto_attach_retry_failed` diagnostics on timeout or failure.
> - Behavioral Change: observation drain no longer reorders rows within a session by attempt count; any row failing a lane head blocks later rows in that lane.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6ea68c6.</sup>
> 
>


TheGaySupreme123:
@coderabbitai full review

PR #295, opened 2026-08-17
Bound the hook budget, stream partials, and store-stage cost

### LINKED ISSUE #288: [bug]: _HOOK_TOTAL_BUDGET_SECONDS is smaller than the budgets nested inside one pass — hook_budget_exceeded fires on healthy hooks (253/833)
## Summary

`_HOOK_TOTAL_BUDGET_SECONDS` is 1.0s, but the budgets nested inside one pass sum to more than that before any work happens. Every hook that reaches a real drain attempt reports `hook_budget_exceeded`, so the diagnostic fires on healthy passes and carries no signal.

## The arithmetic

`src/yoetz/cli/observe_hooks.py`:

- `:110` `_HOOK_DRAIN_BUDGET_SECONDS = 0.20`
- `:126` `_HOOK_CONNECT_PREFLIGHT_SECONDS = 1.0`
- `:137` `_HOOK_TOTAL_BUDGET_SECONDS = 1.0`

The connect preflight alone equals the entire end-to-end budget. `:787-791` starts the drain clock *after* the connect — deliberate, and correct for the drain's own accounting — so a cold drain costs up to 1.0s preflight + 0.2s drain + encode before the `store` stage is even counted.

`_record_pass_timing` (`:614-637`) compares the whole pass against `_HOOK_TOTAL_BUDGET_SECONDS` and records `hook_budget_exceeded` when it is over. Overrun is therefore structural.

## Measured, this workspace

`yoetz observe status`: 833 diagnostics, **`hook_budget_exceeded: 253`**, `drain_lease_contended: 153`, `drain_budget_exhausted: 119`, `runtime_gate_unsafe: 108`. `timings: {count: 257, last_ms: 2065, max_ms: 4453}`.

Timing rows from a single live Codex session:

```
PreToolUse  2065ms  {import: 17, store: 520, drain: 1312, advice: 7}
PostToolUse 1232ms  {import: 20, store: 500, drain:   23, advice: 6}
PreToolUse  2076ms  {import: 25, store: 519, drain: 1302, advice: 6}
PreToolUse  2124ms  {import: 26, store: 563, drain: 1300, advice: 6}
```

The ~1300ms drains are 6.5× the declared 0.20s drain budget, matching `preflight(1.0) + drain(0.2) + encode` exactly. The 23ms drains are the fast bail from `drain_lease_contended` (`:731`). Neither shape can land under a 1.0s total.

## Why this matters even though nothing is killed

`_record_pass_timing`'s own docstring (`:623-628`) says the budget is "Observability only: exceeding the budget never aborts a pass, because that would drop ingest." The host budget is `observe_timeout = 10` (`src/yoetz/adapters/integrations/codex_plugin.py:153`) and the worst pass ever recorded here is 4453ms, so no hook is being SIGKILLed.

So the defect is not lost work — it is that the one diagnostic named after a budget overrun fires on ~30% of all passes under normal operation, which makes it useless for detecting the real regressions it was added for (#209/#210/#242). A contract that a healthy pass cannot meet is not a contract.

## Suggested direction

Either raise `_HOOK_TOTAL_BUDGET_SECONDS` to something the nested budgets can actually fit inside (preflight + drain + a realistic store stage), or make it a derived value rather than an independent constant, so the three cannot drift apart again. A latency assertion tying the total to the sum of its parts would keep it honest — `grep -rn "_HOOK_TOTAL_BUDGET_SECONDS" tests/` currently finds no such test.

## Not covered elsewhere

#209/#210/#211/#242/#218 (all closed) were about real kills and real backlog under the pre-#213 constants. This is the post-#213 constant set being internally inconsistent — introduced by the same fix that moved the drain clock after connect and raised preflight from 0.35 to 1.0 without revisiting the total.

Verified against `f5930fc` (main).

### LINKED ISSUE #289: [bug]: stream_partials is unbounded and not in the eviction ladder — 2 entries hold 41% of a state file at 95% of its hard cap
## Summary

`stream_partials` is the only unbounded collection in the observation state file, and it is also absent from the eviction ladder that `_save` uses when the file hits its cap. On this workspace two entries hold 409 KB — 41% of a file sitting at 95% of the hard 1 MiB limit — so the next overflow will evict envelopes and outbox rows around a blob that cannot itself be shed.

## Evidence

Every other collection in `src/yoetz/adapters/integrations/observation_local.py` is bounded:

```
:82  _MAX_ENVELOPES   = 256
:83  _MAX_DEDUP       = 4_096
:84  _MAX_OPEN_PRE    = 256
:85  _MAX_OUTBOX      = 512
:86  _MAX_QUARANTINE  = 512
:90  _MAX_QUARANTINE_AGE_DAYS = 14
```

`stream_partials` has no counterpart. `set_stream_partial` (`:1310-1319`) stores an arbitrary-length `bytes` value per session commitment with no length check, and it is base64-encoded into the state file at `:2568-2572` / decoded at `:2727-2732`.

The eviction ladder in `_save` (`:2231-2250`) sheds, in order, `envelopes` (noting `TRUNCATED_PAYLOAD`) and then `pending_outbox` rows (folding them into `quarantine`). `stream_partials` is not in that ladder at any position.

## Measured, this workspace

State file `observation/workspaces/722a447…json` is **996,510 bytes** against `_MAX_STATE_BYTES = 1_048_576` (`:80`) — **95.0% of the cap**. Composition:

```
stream_partials   409,092 bytes   n=2      <-- 41.1%
quarantine        248,232 bytes   n=318
envelopes         167,699 bytes   n=256    (at _MAX_ENVELOPES)
dedup             102,600 bytes   n=1368
pending_outbox     42,031 bytes   n=54
session_advice     26,592 bytes   n=12
```

Two entries out of ~2,000 hold more bytes than the 318-entry quarantine. Base64 inflation means the underlying partials are roughly 307 KB of raw stream data held resident.

## Impact

Once the file crosses 1 MiB, `_save` starts deleting real observations — envelopes first, then outbox rows — while the two stream partials are retained untouched. That is durable observation loss caused by a collection that has no reason to be the largest thing in the file, and no mechanism to shrink.

`yoetz observe status` on this workspace also reports `stream_coverage: False`, so the partials are consuming the budget without the stream path being covered at all.

## Suggested direction

Add a `_MAX_STREAM_PARTIAL_BYTES` cap enforced in `set_stream_partial`, and place `stream_partials` in the `_save` eviction ladder ahead of `envelopes` — a partial is by definition resumable-or-discardable, whereas an evicted envelope is a lost observation. A partial that exceeds the cap should drop with an explicit gap code rather than displace durable rows.

## Not covered elsewhere

#211 (closed) bounded the quarantine and added `quarantined_at` + the 14-day age fence; it did not touch `stream_partials`. Nothing in the open or closed set mentions stream partials or state-file composition.

Verified against `f5930fc` (main), with live state from workspace commitment `hmac-sha256:722a447…`.

### LINKED ISSUE #290: [bug]: hook store stage costs ~500ms on every pass and is not attributable — JSON parse/encode of the same state is 5ms
## Summary

The `store` stage of every hook pass costs ~500 ms, consistently, on every event. It is the single largest fixed cost in a hook and it is paid whether or not anything is delivered. Serialization is not the cause: parsing and re-encoding the entire state file measures 5 ms.

## Measured, live Codex session

Timing rows from `observation/hook-diagnostics.jsonl`:

```
PreToolUse  2065ms  {import: 17, store: 520, drain: 1312, advice: 7}
PostToolUse 1232ms  {import: 20, store: 500, drain:   23, advice: 6}
PreToolUse  2076ms  {import: 25, store: 519, drain: 1302, advice: 6}
PreToolUse  2124ms  {import: 26, store: 563, drain: 1300, advice: 6}
PostToolUse 1528ms  {import: 22, store: 492, drain:  23, advice: 6}
```

`store` is 492–563 ms across every row, independent of event type and independent of whether the drain got the lease (1300 ms) or bailed on contention (23 ms).

## Serialization is not where it goes

Against the same live state file (996,510 bytes):

```
json.loads  →  2.1 ms
json.dumps  →  3.0 ms
```

So ≥99% of the 500 ms is somewhere other than JSON I/O — the per-entry hydration in `_state_from_json` / `_load_*` (`src/yoetz/adapters/integrations/observation_local.py:2727-2923`), the interprocess store lock (`:151-203`, `_STORE_LOCK_TIMEOUT_SECONDS`), the canonical re-encode in `_save` (`:2227-2230`), or the atomic write. The stage is not currently broken down further, so the diagnostic cannot say which.

The collections being hydrated on every pass: 318 quarantine, 256 envelopes, 1368 dedup, 54 outbox, 12 session_advice, 19 cursors.

## Why it matters

A hook is always a fresh process (`observe_hooks.py:121-125`), so this is paid in full on every PreToolUse and PostToolUse — the highest-frequency events in a session. At ~500 ms × 2 per tool call it is the dominant term in observation overhead, and it consumes half of `_HOOK_TOTAL_BUDGET_SECONDS = 1.0` (`:137`) before the drain clock even starts. See the companion issue on the budget arithmetic.

PR #213 fixed the *repeat*-parse problem (17 parses per invocation → 1, via the stat-validated bounded parse cache). The remaining single hydration is what this issue is about.

## Suggested direction

First, split the `store` stage in `_record_pass_timing` (`:614-637`) into hydrate / mutate / encode / write sub-stages so the cost is attributable — right now the diagnostic reports a number nobody can act on. Then bound whatever dominates. A plausible candidate is that the full state is hydrated for events that only append one envelope and never read quarantine, dedup history, or session advice.

Any fix needs a latency assertion against a realistically-sized store; a fixture-sized store passes trivially and would not have caught this.

## Not covered elsewhere

#209/#210/#242 (closed) covered the pre-#213 catalog handshake and sync-hook kill loop. #218 (closed) covered drain throughput. None of them measured or bounded the store stage itself.

Verified against `f5930fc` (main), with live telemetry from workspace commitment `hmac-sha256:722a447…`.

TheGaySupreme123:
Fixes three defects that all land in the same hook pass, all introduced or left behind by the post-#213 constant set.

Closes #288
Closes #289
Closes #290

## #288 — the total budget was smaller than its own parts

`_HOOK_TOTAL_BUDGET_SECONDS` was `1.0`, but the budgets enforced sequentially inside one pass already summed past it: the connect preflight alone is `1.0`, before the drain budget (`0.20`) or any local work. Overrun was structural, so `hook_budget_exceeded` fired on **253 of 833** diagnostics on a live workspace and carried no signal about the regressions (#209/#210/#242) it was added to catch.

The total is now **derived** rather than an independent constant:

```
_HOOK_TOTAL_BUDGET_SECONDS = preflight (1.0) + drain (0.20) + local-stage allowance (1.0)
```

A test asserts the derivation and that the two measured healthy shapes (cold drain ~2.1s, contended bail ~1.3s) no longer trip the diagnostic, so the parts cannot silently drift past the whole again. Events that may legitimately retry auto-attach (`_AUTO_ATTACH_RETRY_EVENTS` plus `SessionStart`) carry that enforced budget on top; the high-frequency `PreToolUse`/`PostToolUse` storm never attaches and keeps the tighter contract.

## #289 — `stream_partials` was unbounded and outside the eviction ladder

It was the only unbounded collection in the state file *and* absent from the `_save` ladder, so two entries held 409 KB — 41% of a file at 95% of the hard 1 MiB cap — while overflow evicted envelopes and outbox rows *around* a blob that could not itself be shed.

A partial is a pure read-cache: the reader seeks to the committed cursor and rereads the tail whenever the partial is absent, so dropping one costs a bounded reread, never observation loss. Accordingly:

- bounded per entry at 64 KiB;
- an oversized tail now **drops with an explicit gap instead of raising** — the previous `ProtocolValueError` stalled the stream forever while retaining the very partial that caused the stall;
- shed **largest-first, ahead of envelopes and outbox rows** in the ladder;
- projects as `source_lag` on status (the source tail is pending a reread), never leaking the internal marker;
- states persisted before the bound existed are healed by the next save.

## #290 — the ~500 ms store stage, attributed and then removed

Profiling the live 1 MiB state showed the cost was not JSON I/O but the canonical codec: `_validate_string` and `_encode_string` walked **every character in a Python loop**, running ~70× the stdlib on the same bytes. Both hot loops now take a single C-level regex scan and fall back to the exact original logic only when a string actually contains an escape or an invalid codepoint. `_decode_object_pairs` now uses the C `dict()` constructor with a length check for the duplicate-key case.

Measured on the live 1 MiB state file, best-of-5:

| stage | before | after |
|---|---|---|
| `strict_json_parse` | 81.9 ms | 12.7 ms |
| `_state_from_json` | 37.2 ms | 17.1 ms |
| `_state_to_json` + encode | 155.2 ms | 24.3 ms |
| `_atomic_write` (fsync) | 0.5 ms | 0.5 ms |
| **total** | **274.8 ms** | **54.6 ms** |

Timing rows also now break `store` into `store_hydrate` / `store_encode` / `store_write`, so any remaining cost is attributable rather than a number nobody can act on.

### Equivalence of the codec change

The codec is load-bearing for digests and wire identity, so the fast paths were **differentially fuzzed against the pre-change implementations** over 539,268 comparisons — 200k random strings drawn from a boundary-weighted alphabet, plus an exhaustive sweep of every BMP codepoint and several astral ones. Output and raised-error identity are byte-identical in every case. The escaped-character class is a strict superset of the invalid-character class, so a non-match proves both "nothing to escape" and "nothing to reject".

## Verification

- `3515 passed, 15 skipped` across the suite (packaging excluded — it requires a non-worktree checkout).
- Full-repo Pyright: `0 errors, 0 warnings`; Ruff format + check clean.
- New: `tests/unit/protocol/test_canonical_performance.py` (latency fence against stdlib on a ~1 MiB document — a fixture-sized store would pass any implementation trivially), `tests/unit/adapters/test_observation_state_bounds.py`, and three cases in `tests/unit/cli/test_observe_hooks.py`.

> [!NOTE]
> ### Bound stream partial size, derive hook budget, and attribute store sub-stage timings
> - Stream partials in `LocalObservationStore` are now bounded at 262 144 bytes per entry; oversized partials are dropped and the session is marked, projecting a `SOURCE_LAG` gap until recovery. The gap resolves automatically when all affected sessions restart or end.
> - `SessionStreamReader.advance` performs bounded read-ahead (up to `max_line_bytes + 1`) when no newline is present, improving assembly of long lines across chunks.
> - The end-to-end hook budget in `observe_hooks` is now derived from enforced sub-budgets plus a local-stage allowance rather than a fixed constant; `SessionStart` and auto-attach events carry an additional retry allowance.
> - `LocalObservationStore` now accumulates monotonic timings for hydrate, encode, and write sub-stages; these are attributed as `store_hydrate`, `store_encode`, and `store_write` in timing rows recorded by `handle_observe`.
> - `canonical.py` string validation and encoding gain regex-based fast paths that skip per-character loops for plain ASCII strings; strict JSON parse uses length comparison for duplicate key detection.
> - Risk: persisted state schema version bumps from 6 to 7; older readers expecting schema `/6` will reject state files written by this version.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 991a932.</sup>
> 
>


TheGaySupreme123:
Fixed in 991a932: the reader conditionally reads ahead to the profile's maximum line size plus its terminator only when the ordinary chunk has no newline, so a dropped live partial recovers when the newline arrives without raising the persisted partial cap. Regression coverage was added.

PR #671, opened 2026-09-09
fix(hosts): ingest host work into the ledger correctly across Codex, Claude Code, and Cursor

### LINKED ISSUE #619: fix(observation): schedule semantic advice asynchronously and preserve coverage gaps
# Enable observation semantic advice as bounded asynchronous work

## Summary

Observation semantic advice is wired into the READY composition but is currently dormant, and
the same callback drops the observation coverage gaps that a future implementation would need to
send to a provider.

## Evidence

In `src/yoetz/service/ready_composition.py`, `provide_service_ready_context` creates
`ObservationAdviceContextBuilder(semantic_review=_semantic_review)` when semantic review is
configured. The closure captures `provider_binding = None` and `semantic_ready = False`; neither
value is assigned later. `_semantic_review` therefore returns before constructing a provider
candidate on every invocation. It also discards its `gaps` argument and currently builds the
minimized packet with `coverage_gaps=()`.

`ObservationAdviceContextBuilder.build` in `src/yoetz/application/observation_advice.py` invokes
the callback with the scoped status gaps and awaits an async result. The observation coordinator
calls `_run_advice` after hook ingest and from verification-drain/recovery paths, so making the
callback call the provider directly would put a potentially 60-second privacy/model operation on
each advice update. That would turn a dormant path into hook latency and duplicate-dispatch risk.

## Relationship to existing work

- Closed issue #13 described the original asynchronous privacy-gated observation semantic advice
  feature and should be treated as the design source.
- #571 covers ordinary check latency and isolation behavior; it does not cover this callback's
  scheduling or deduplication.
- #616 covers native Cursor integration reliability; it is related to the host path but does not
  resolve semantic advice scheduling.
- #618 covers guidance and should be updated only after the runtime behavior and recovery rules
  are defined.

## Required fix

Design and implement a bounded asynchronous observation-advice semantic worker before activating
the callback. The worker should:

1. resolve repository-scoped provider authority at dispatch time, rather than using the static
   READY snapshot;
2. preserve the complete bounded observation gap set in the evidence packet and its basis digest;
3. enqueue or coalesce work outside the hook's critical path, with a finite queue and explicit
   pending/unavailable coverage when work cannot run;
4. deduplicate by the advice evidence-basis identity and session/task route so repeated hook
   updates do not mint repeated provider attempts;
5. persist and recover attempt state across daemon restart without treating an unattempted review
   as successful; and
6. keep provider authorization, receipts, semantic output validation, and coverage degradation
   independently visible in the advice and receipt surfaces.

## Acceptance criteria

- A configured provider and an authorized repository cause a semantic advice attempt through a
  worker, with no provider call on the hook's synchronous path.
- Repeated identical advice builds produce at most one in-flight or durable attempt for the same
  basis/session identity.
- A changed basis schedules a new attempt and retains the prior attempt receipt and coverage
  history.
- The provider packet contains the exact bounded observation gaps supplied by the scoped status.
- Queue full, provider unavailable, authorization missing, cancellation, and restart each leave a
  truthful bounded gap and no semantic success claim.
- Tests prove the callback cadence, coalescing, restart recovery, gap preservation, and receipt
  accounting without live provider calls.

### LINKED ISSUE #656: [change]: keep compatible host upgrades working without per-release allowlisting
Priority: **P1 — product compatibility**. New host releases repeatedly disable a useful integration capability until Yoetz publishes a release-specific profile. No security bypass, data corruption, or universal MCP outage is alleged.

## Duplicate search

- Searched all issues and PRs, plus targeted compatibility/profile searches on 2026-09-08.
- [#568](https://github.com/TheGaySupreme123/yoetz/issues/568) / [PR #583](https://github.com/TheGaySupreme123/yoetz/pull/583) added the exact 0.150.1 parser profile. This proposal replaces that ongoing per-release requirement rather than adding 0.153.4 as another exception.
- [#413](https://github.com/TheGaySupreme123/yoetz/issues/413) establishes exact capability evidence; [#616](https://github.com/TheGaySupreme123/yoetz/issues/616) / [PR #617](https://github.com/TheGaySupreme123/yoetz/pull/617) repair observation and responsiveness. Neither owns version-independent transcript admission.

## Problem statement and reproduction

Observed in the independent Yoetz 0.1.0 installation built from `fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe` on macOS, with Codex host 0.153.4. The separate semantic evaluator was Codex 0.150.1. Evidence comes from the 2026-09-08 issue-654 dogfood; [PR #655](https://github.com/TheGaySupreme123/yoetz/pull/655) is the coding task, not an installation of a repaired parser.

The installed `SUPPORTED_ROLLOUT_PROFILES` contains only 0.148.0 and 0.150.1. A transcript header declaring 0.153.4 is rejected by exact-key admission even when its visible events are structurally compatible with a supported grammar. Final diagnostics reported `unsupported_format` and `codex_session_stream=false`, while hooks retained 73 immutable snapshots and semantic checks worked through the separate evaluator.

Reproduce the gating problem using a privacy-safe supported transcript fixture: preserve its event structure, change only its host-version field to a new release, and run parser/reconciler admission. The version alone currently prevents the stream from being admitted. Separately validate actual 0.153.4 event families; this issue does not assume every event in every future release is compatible.

## Why it happened: exact release identity is being used as format compatibility

The Codex failure is a deterministic admission decision, not an unexplained parser crash:

1. [The installed profile table](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/adapters/importers/codex_rollout_jsonl.py#L153) has two exact keys, `0.148.0` and `0.150.1`.
2. [`_admit_session_meta`](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/adapters/importers/codex_rollout_jsonl.py#L433) reads `payload.cli_version` and looks it up directly. An absent key returns `unsupported_codex_profile` before a compatible event body can establish support. It also checks the history mode; that is a separate structural constraint and should remain meaningful.
3. [The session-stream reader](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/adapters/integrations/codex_session_stream.py#L701) preserves refused admission for the current source generation. Later appends do not recover admission: with consumed bytes but zero admitted events, it returns no normalized events and `unsupported_format`. Rotation/truncation opens a new generation, but the unchanged unsupported header would still fail the same gate.
4. Consequently, ordinary native hooks can capture objects and the separately configured evaluator can complete semantic review while the transcript reconciliation lane remains unavailable. Those successes do not repair missing stream-derived outcomes or pairing.

A read-only function probe in the **installed runtime** held the header shape and accepted history mode constant and changed only `cli_version`. `0.150.1` admitted with `session_meta`; `0.153.4` and synthetic `99.0.0` both refused with `unsupported_codex_profile`. This proves the version gate independently of the real transcript's other format differences. It is not a claim that the complete 0.153.4 stream has already passed structural conformance.

The architectural cause is coupling an exact-release certification claim to baseline useful operation. Certification may legitimately describe versions exercised in a live test. Compatible event interpretation should not require a new certification entry for every patch release.

## Is this only Codex? Host-by-host scope

| Host | Code-traced behavior at the audited revision | Evidence and required treatment |
| --- | --- | --- |
| Codex | Exact transcript-header gate rejects new versions; rejected generation stays unadmitted. | Observed live on 0.153.4 and reproduced with an identical-shape header probe. Primary incident. |
| Claude Code | Native plugin preview/apply preparation rejects a target version outside the one-element `supported_versions` list; hook version mapping also maps unfamiliar versions to `untested`. | Source-confirmed additional upgrade gate. No new live Claude upgrade was run in this audit. Include native installation/admission and hook compatibility in the design. |
| Cursor | Hook version mapping uses an exact table; a supplied unknown version maps to `untested` and takes the conservative pairing contract. An omitted version takes a separate compatibility path. | Source-confirmed qualification/fallback distinction. This is not evidence that all Cursor hooks or MCP calls stop. Test what capabilities are lost and which compatible events continue. |

Claude owners: [native plugin admission](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/adapters/integrations/claude_code_integration.py#L1641) and [hook version mapping](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/cli/observe_hooks.py#L3262). Cursor owner: [hook version mapping](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/cli/observe_hooks.py#L491). These are different host boundaries; the Codex JSONL parser is not used to parse Claude or Cursor events.

The requested design therefore covers **compatible upgrades of all three hosts**, with separate structural contracts per host/surface. Keep tested-version provenance separate from the decision to perform already-understood operations. Preserve consent, identity, content exclusions and truthful per-capability limits. Unknown structure must not inherit trusted outcome semantics merely because another host or release used a similar name.

## Concrete upgrade regression matrix

For each supported host surface, run an unchanged valid payload under its existing version, a previously unseen version, and an omitted version where that host permits omission. Record installation/admission, event ingestion, pairing/outcome interpretation, capture and status independently. Add an incompatible payload under a familiar version: a known version must not make invalid structure acceptable.

For Claude, exercise plugin preview and the actual installed entrypoint in a disposable instance, not only the hook mapper. For Cursor, exercise both omitted and supplied-unknown version branches and distinguish IDE from CLI capability scope. For Codex, test append/restart after rejected admission so upgrading Yoetz or changing compatibility policy cannot silently strand a persisted cursor. Define bounded replay/re-admission behavior without duplicate events or loss of the refused range.

Routine compatible upgrades should require neither manual per-release approval nor a new Yoetz allowlist release. Arbitrary breaking changes still require an adapter change. This proposal does not broaden semantic evaluator execution or disclosure authority as a side effect of host compatibility.

## Proposed approach

Admit and normalize transcript events based on a versioned structural contract and supported capabilities. Host version remains diagnostic provenance, not a per-release parsing gate. Compatible new releases should work automatically, without manual approval or a new Yoetz release.

Recognize stable wrapper/event families; ignore irrelevant additive fields before retention; classify incompatible or unknown event structures as bounded, source-specific coverage gaps while processing independent supported events. Preserve pairing identities, byte/content commitments, pagination/compaction handling, partial-tail recovery, cursor durability and idempotency. If an unknown event could affect a cross-event conclusion, weaken that conclusion explicitly.

## Affected area and design gate

[Parser](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/adapters/importers/codex_rollout_jsonl.py), session-stream reconciler, capability/parity reporting, fixtures and [ADR-005](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/docs/adr/ADR-005-codex-capability-identity.md) / [ADR-010](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/docs/adr/ADR-010-harness-integration-port.md).

**Design-gated: yes.** ADR-005 explicitly requires exact-version profiles today. The requested new behavior needs an explicit amended contract before implementation sign-off, rather than an undocumented bypass. The maintainer requested automatic compatibility in the dogfood discussion; this issue records the concrete proposal for issue-level acknowledgement.

## Acceptance criteria

- [ ] Changing only a supported fixture's version to a previously unseen release does not disable compatible events or require another allowlist entry.
- [ ] Actual 0.153.4 fixtures cover the observed wrappers, tool calls/results, pagination, world state, compaction, and incomplete tails; older supported shapes retain their semantics.
- [ ] Additive fields do not break compatible events or enter retained content unintentionally.
- [ ] Malformed/incompatible known events and unknown event types produce bounded diagnostics; unrelated valid events continue without silent cursor loss, duplicate publication, invented outcomes or guessed pairing.
- [ ] Hidden reasoning, private instructions, credentials and unrelated data remain excluded. Parser compatibility grants no capture consent or egress authority.
- [ ] Status distinguishes structurally supported, partially understood and incompatible streams and names the affected event family. Parsing alone does not certify every host feature.
- [ ] Installed Codex-testing dogfood proves real stream reconciliation, outcome/pairing facts, compaction and terminal drain using the new admission policy.
- [ ] ADRs, interfaces, fixtures, generated support metadata and parity gates agree on the new scope.

## Non-goals

Guaranteeing correct interpretation of an arbitrary incompatible future format; aliasing every unknown version to the latest exact profile; widening content consent; changing the independently bound semantic-evaluator executable or its disclosure authority.

## Required compatibility boundaries

Build a differential fixture matrix: existing supported shape with a new version label; harmless extra fields; unknown independent event; incompatible known-event payload; conflicting/missing correlation identity; compaction and replacement-history transition; and a truncated final JSONL line. Separate successful parsing of an event from whole-session coverage. An unknown event that could change an outcome or correlation must weaken affected conclusions rather than be silently ignored.

Do not make receipt or native-host support labels broader merely because the parser accepts a future version. The practical acceptance target is that routine compatible Codex upgrades keep useful observation working without a release-specific approval loop, while actual format changes remain diagnosable.

## Evidence provenance

Prepared by GPT-6 Astra in Codex desktop from a read-only audit of the completed Codex-testing run. The report used host-recorded MCP results, supported projected status/history, installed-source inspection and final remote PR state. No product code was changed and no live workflow writes were fabricated to obtain these findings. Observed facts, proposed reproductions and unproven causes are distinguished above.

### LINKED ISSUE #657: [bug]: receipt describes hook-only frontier movement as finding responses
Priority: **P2 — incorrect receipt explanation and misleading retry advice**. The check remains attributable and receipt creation succeeds. No false clean verdict or retry livelock was observed.

## What actually happened

At the end of the 2026-09-08 Codex-testing dogfood, the agent ran a final check and then requested a receipt. Between those operations, Yoetz's automatic observation writer appended records describing host tool activity. The agent did **not** answer any findings in that interval. The receipt nevertheless claimed that only finding responses had been published.

Observed installation: Yoetz 0.1.0 built from `fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe`, macOS, Codex host 0.153.4, separate semantic evaluator Codex 0.150.1. The coding work was [PR #655](https://github.com/TheGaySupreme123/yoetz/pull/655); that PR did not replace the checker installed for this audit.

“Frontier” here means the ledger's ingestion sequence: the number identifies the accepted history prefix, not a Git commit and not the time an observed tool originally ran.

| Step | Accepted ledger sequence | Meaning |
| --- | --- | --- |
| Final check input | 540 | The history the check actually evaluated. |
| Check recorded | 541 | The completed check was durably recorded at 12:01:23.908 UTC. |
| Automatic observations | 542–549 | Five `action_recorded` and three `result_recorded` events, all `hook_observed`, actor `yoetz:observation-coordinator`. Zero `response_recorded` and zero new findings. |
| Receipt input | 549 | The receipt includes the later observations while retaining the check's earlier tested boundary. |

The first three observation rows in that suffix describe activity at 12:00:22–12:00:23, before the final check was recorded, but were accepted after it. This is asynchronous delivery, so a later ingestion sequence is not itself proof that the agent did new cooperative work after review.

The complete supported history view at frontier 549, filtered after 540, returned no continuation cursor:

| Sequence | Record | Channel |
| --- | --- | --- |
| 541 | check_recorded | engine_derived |
| 542–543 | action_recorded | hook_observed |
| 544 | result_recorded | hook_observed |
| 545 | action_recorded | hook_observed |
| 546 | result_recorded | hook_observed |
| 547–548 | action_recorded | hook_observed |
| 549 | result_recorded | hook_observed |

Receipt `rcp_5899a294-6fff-461f-a430-692bfc65a4db` said:

> A check is recorded at subject frontier 540 and still contributes here: only responses to the findings it returned were published after it.

It then stated that the verdict covers 540, not 549, and instructed the reader to rerun the check at this frontier. The earlier frontier statement is accurate; the claimed response history is false.

## Why it happened: the applicability rule expanded, but the explanation did not

This is a confirmed mismatch between shared decision logic and shared rendering, not an unknown host failure:

1. [`_record_supersedes_recorded_check`](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/kernel/reducers.py#L160) first checks service-stamped observation authorship. An observation record invalidates the old check only if it is `finding_recorded`. Finding-free observation actions/results therefore leave the check attributable. Cooperative material publications still invalidate it through the ordinary rule.
2. [Receipt `_context`](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/application/receipt.py#L461) uses that rule and finds no invalidating event. It retains `applicable_check`.
3. In that same branch, it sees later events from material event families (`action_recorded`/`result_recorded` qualify by family) and adds `check_current_as_of_earlier_frontier`. The source comment explicitly says that both check-answering responses **and finding-free observation-authored records** can reach this branch.
4. [`_check_coverage_sentence`](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/kernel/receipt_builder.py#L746) receives only gap codes, the receipt frontier, and the tested subject sequence. It receives no classification of the intervening events. Whenever this gap is present, it unconditionally emits “only responses to the findings it returned were published after it.”
5. [The existing renderer test](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/tests/unit/kernel/test_receipt_builder.py#L467) asserts that exact response-only sentence for the general gap. Thus the test protects the obsolete wording instead of distinguishing the two valid reasons for earlier-check attribution.

The shared gap describes **an attributable earlier check**, not a response-only suffix. The renderer incorrectly treats those as equivalent. A read-only direct call to this function in the audited installed runtime reproduced the exact incorrect 540/549 sentence, with no service call or ledger mutation.

Related [#361](https://github.com/TheGaySupreme123/yoetz/issues/361) / [PR #363](https://github.com/TheGaySupreme123/yoetz/pull/363) added observation-aware applicability. [#320](https://github.com/TheGaySupreme123/yoetz/issues/320) / [PR #328](https://github.com/TheGaySupreme123/yoetz/pull/328) address acquisition races. Those mechanisms work in this run; this issue owns the stale explanatory assumption and retry advice.

## Host and delivery scope

**Shared Yoetz defect: Codex, Claude Code, Cursor, or another caller can encounter it when their ledger contains the same accepted suffix.** The function has no Codex-version or host check. Codex is the live reproduction, not the scope of the renderer bug. No separate live Claude/Cursor reproduction is claimed.

Check all receipt surfaces that use the generated sections/human text: MCP and CLI delivery, markdown/text, and structured receipts containing the same sections. The explanation must agree with the applicability decision regardless of delivery path. [Receipt capacity/preflight](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/kernel/receipt_capacity.py#L171) also recognizes the same general gap; preserve its parity if context or gap representation changes.

## Expected fix and advice

The minimum correction is factual wording that does not invent an event class. A more useful explanation derives a bounded suffix classification and names observations, responses, or their mixture when established. For this incident, suitable semantics are:

> The check evaluated frontier 540 and remains attributable. Later records through frontier 549 are finding-free host observations. They are retained in this receipt but were not evaluated by that check.

If the implementation cannot establish the exact suffix class, use neutral wording rather than guessing. Keep the checked boundary explicit. Explain that a new check is needed **to evaluate later material**, and that routine asynchronous observation can advance the ledger again. Do not imply that repeatedly chasing the latest integer frontier is necessary to retain an already attributable check or will necessarily remove every coverage gap.

The terminal conclusion `unresolved_findings_remain` concerns a separately retained semantic finding, tracked in [#660](https://github.com/TheGaySupreme123/yoetz/issues/660). Fixing this sentence must not resolve that finding, erase observation gaps, or strengthen coverage.

## Regression acceptance

- [ ] No suffix: no invented later activity or earlier-frontier explanation.
- [ ] Responses to the checked findings only: truthful response explanation and preserved tested boundary.
- [ ] Finding-free service-stamped observations only: truthful observation explanation; no response-only sentence.
- [ ] Allowed mixed suffix: disclose the mixture or use accurate neutral wording.
- [ ] Observation-authored finding: old check is invalidated as today.
- [ ] Cooperative action/result/claim/evidence or response to another finding: material supersession still applies; an observation-shaped but unauthenticated payload cannot gain the exception.
- [ ] Late-delivered records and active observation writer: ingestion order remains distinct from occurrence time; advice does not require an unbounded retry loop.
- [ ] Shared application, renderer and capacity tests agree, including memory/SQLite projection and replay where affected. Use host-shaped inputs from all three adapters for the shared authorship boundary.

Design gate: factual wording can preserve existing semantics. New public suffix fields, gap codes or applicability rules need an explicit contract decision and the corresponding docs/schema update.

## Evidence limits

Evidence is a completed native Codex run, complete bounded projected history, the recorded receipt, and exact installed-source/function inspection. The linked product source was checked against both the immutable revision and installed package. This audit did not rerun the workflow, reproduce a livelock, or test a fix. Prepared by Codex for the maintainer-requested issue expansion.

### LINKED ISSUE #659: [bug]: Codex compaction hook still probes a mapped task without workspace context
Priority: **P2**. The fallback worked and no mapping was lost; automatic repository verification still failed at a normal compaction.

## Duplicate search

Related closed [#578](https://github.com/TheGaySupreme123/yoetz/issues/578), fixed by [PR #589](https://github.com/TheGaySupreme123/yoetz/pull/589). Its merge commit `57fa782a5f6149c7367013355160bd28609c0568` is an ancestor of the observed installation commit. This is a residual caller-path occurrence after that repair. Unlike the original report, the hook now correctly preserves the mapping and does not call it stale. Do not claim the original stale-mapping defect is back.

## Area and version

Codex mapped-session `SessionStart`/compaction hook, workspace locator propagation and repository-fenced status probe. Observed in the independent Yoetz 0.1.0 installation built from `fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe` on macOS, with Codex host 0.153.4. The separate semantic evaluator was Codex 0.150.1. Evidence comes from the 2026-09-08 issue-654 dogfood; [PR #655](https://github.com/TheGaySupreme123/yoetz/pull/655) is the coding task, not an installation of a repaired parser.

## Reproduction evidence

1. Start and map a repository-bound task in an isolated Codex-testing host.
2. Continue substantive work until a natural context compaction.
3. At 11:58:33 UTC, the run compacted. The injected hook message reported `status_workspace_unbound`: it could not verify the repository because the hook connected without a project workspace.
4. The message correctly said the mapping was not stale, the service was healthy, and not to reattach.
5. A subsequent model-issued status call from the project recovered the same task/session/writer at frontier 525.

The repository-context failure is code-traced below. The audit also identifies two caller paths that can produce it and the exact missing diagnostic needed to tell them apart in this incident.

## Why the advisory was produced: repository fencing, not a lost task

The failure layer is confirmed. A hook had the task/session/writer mapping but its status request reached the task workflow without an established repository privacy context:

1. [The service handshake](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/service/control_protocol.py#L865) starts with `repository_context=None` and resolves a supplied workspace locator when a resolver is available.
2. [The task repository fence](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/application/service.py#L782) compares the request's repository commitment with the stored attachment. If the request has no context, it rejects with `SESSION_CONFLICT` and the specific reason `repository_identity_required`.
3. [Hook failure classification](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/cli/hooks.py#L229) maps that reason to `workspace_unbound`, rather than treating every session conflict as a stale mapping.
4. The hook emits the observed advisory and preserves the mapping. A later model-issued project-bound status request establishes the context and reads the same task/session/writer at frontier 525.

Therefore this incident is an automatic status probe lacking **usable repository context**. The advice string says “without a project workspace,” but that alone cannot distinguish an absent locator from a supplied locator that failed to resolve to repository context. The task/session IDs by themselves are deliberately insufficient to bypass the repository fence.

## Code-traced caller weaknesses and the remaining evidence boundary

The installed source contains two relevant SessionStart paths:

| Path | Where workspace comes from | Failure mode to reproduce |
| --- | --- | --- |
| Codex resume/compact `hooks session-start` | The renderer omits `--workspace`; `session_workspace_locator()` defaults to process `.`. Its caller does not pass host payload workspace/cwd fields to that resolver. | A hook launched outside the repository, or a rejected canonical path, does not bind the intended project even if the host payload contains a valid project location. |
| Shared `hooks observe` mapped SessionStart | An explicit consented workspace yields a canonical locator. With no explicit workspace, a legacy fallback can recover a workspace commitment by session while leaving `workspace_locator=None`. | Mapping/consent lookup can succeed, then the mapped status probe uses an unbound connector. A commitment used to find local state does not reconstruct the path needed for the service handshake. |

Sources: [Codex renderer, both SessionStart commands](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/adapters/integrations/codex_plugin.py#L165), [cwd-only session locator](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/cli/hooks.py#L646), [mapped session caller](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/cli/hooks.py#L880), [shared observation workspace selection](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/cli/observe_hooks.py#L2422), and [shared mapped status connector](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/cli/observe_hooks.py#L2900). The source owner for bounded host-root selection already states that Codex/Cursor hook working directories are not stable across surfaces: [workspace binding module](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/adapters/workspace_binding.py#L1).

This narrows the implementation problem to deriving and carrying repository context consistently across entrypoints. Adding `--workspace .` alone is insufficient if the underlying hook cwd is wrong. Likewise, suppressing the warning would hide the failed verification rather than fix it.

**The exact incident's final caller branch is not recorded.** The transcript retained the injected `hooks.additional_context` message at 11:58:34.913 UTC, but not the hook PID, raw SessionStart payload, chosen locator source, hook cwd or handshake resolution result. Both entrypoints can emit the same text. The audit therefore cannot honestly assert which of the code-traced paths fired in that process. This is a specific missing discriminator, not an unspecified root cause or evidence that the service died.

## Is it Codex-only?

- **Codex:** native resume/compaction incident observed; its dedicated re-ground command has the cwd-only selection path described above.
- **Claude Code:** ordinary native rendering passes `--workspace "${CLAUDE_PROJECT_DIR}"` and routes through the shared observation handler. It does not use Codex's dedicated `session-start` command. Missing/empty environment context still needs a typed safe failure; no live Claude compaction failure was observed here. See [Claude hook renderer](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/adapters/integrations/claude_code_integration.py#L676).
- **Cursor:** native rendering passes `--workspace .`, while its hook ingress has a bounded resolver for host workspace fields/environment. It then uses the shared observation path. Do not assume the Codex-specific failure automatically reproduces there. See [Cursor hook renderer](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/adapters/integrations/cursor_integration.py#L737) and [Cursor workspace resolver](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/cli/observe_hooks.py#L511).

The repository fence and status classification are shared. The verified incident is Codex-specific; the fix needs cross-host caller regression coverage because shared locator handling can affect all three hosts.

## Reproduction and diagnosis required to close the remaining branch uncertainty

Use an independent installed instance and the **rendered hook commands**, preserving the real mapping. Exercise root cwd, subdirectory cwd, a non-project cwd with a valid host project locator, absent locator, and conflicting project roots. Test resume and natural/forced compaction separately. Also exercise mapped `handle_observe` with an active consented legacy session binding but no explicit workspace.

Record bounded structural diagnostics: entrypoint, locator source (`explicit`, `host_payload`, `cwd`, `none`), canonicalization success/failure class, whether the handshake received a locator, whether repository context resolved, and the typed status outcome. Do not log raw paths, host payloads, or repository commitments. This distinguishes selection failure, propagation failure and repository resolution failure without weakening privacy.

A repair must validate the existing mapping automatically whenever trustworthy project context is available, preserve the same writer/session, and retain the current no-reattach fallback when context is genuinely unavailable. Cross-host acceptance must name the host entrypoint actually exercised, rather than treating a unit test with a supplied workspace as live compaction proof.

## Expected behavior and acceptance criteria

- [ ] Carry a safely derived canonical workspace into the mapped-session status probe when host/project context exists.
- [ ] A real isolated native compaction validates the existing mapping without an unbound-probe advisory; status retains the same task/session/writer.
- [ ] Cover repo root, subdirectory, supported worktree identity, and supported host payload/command variants.
- [ ] Missing or conflicting workspace context remains a typed safe failure; it must not authorize access to a different repository or rotate a valid session.
- [ ] Preserve the successful keep-mapping/no-reattach recovery for truly unbound probes.
- [ ] Record the actually exercised host/version and hook entrypoint in the regression evidence.

Source owners: [hook entrypoint](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/cli/hooks.py), [observation hooks](https://github.com/TheGaySupreme123/yoetz/blob/fc22a2e7dbec418271cd0b8f7bc1d3a79c9756fe/src/yoetz/cli/observe_hooks.py), and the Codex hook renderer. No storage or privacy relaxation is requested; any changed identity contract is design-gated.

## Observed recovery boundary

The native additional-context message was explicit: “The mapping is not stale and the service is healthy; do not re-attach. Call status from the project directory before further material work.” The next project-bound status call succeeded and reported the existing closed obligation and unresolved finding. The confirmed failure is missing usable repository context on the hook probe. Its remaining uncertainty is the locator-selection/resolution branch, not task or writer loss.

The regression should exercise the actual rendered native hook invocation around compaction, not only `_read_status` with a workspace supplied by a test. Record whether context came from the host event, explicit `--workspace`, or validated hook working directory, and verify that no broader repository fallback is introduced when all are absent.

## Evidence provenance

Prepared by GPT-6 Astra in Codex desktop from a read-only audit of the completed Codex-testing run. The report used host-recorded MCP results, supported projected status/history, installed-source inspection and final remote PR state. No product code was changed and no live workflow writes were fabricated to obtain these findings. Observed facts, proposed reproductions and unproven causes are distinguished above.

TheGaySupreme123:
## Issue

- Linked issue: Fixes #657, Fixes #659, Fixes #619, Fixes #656
- I searched existing issues and PRs for duplicates before opening this PR.
- Design-gated areas (#619 storage/durability, #656 ADR amendment): the maintainer requested this scoped work directly on 2026-09-08 and acknowledged #656 on the issue ("ack"); the #619 acknowledgement is recorded in an issue comment.

## Documentation

- Behavior change? `yes`
- Docs updated: `docs/adr/ADR-005` (structural admission amendment), `docs/adr/ADR-010` (compatible host upgrade amendment), `docs/INTERFACES.md` (receipt suffix classes, locator source diagnostics, async advice worker, Codex admission states, Claude floor admission), `docs/runbooks/{codex,claude-code,cursor}-integration.md`, `docs/usage/receipts-and-coverage.md`, `guidance/coverage-and-receipts.md` (+ generated mirrors).

## Verification

Commands run (paste):

```text
uv run pytest tests/unit/kernel/test_receipt_builder.py tests/integration/application/test_respond_status_receipt.py -q        # 82 passed
uv run pytest tests/unit/cli/test_hooks.py tests/unit/cli/test_observe_hooks.py tests/unit/cli/test_hook_diagnostics.py \
  tests/unit/adapters/test_codex_plugin.py tests/unit/cli/test_plugin_command_matrix.py -q                                    # 285 passed
uv run pytest tests/unit/application/test_observation_advice_semantic.py tests/unit/application/test_observation_advice.py \
  tests/unit/application/test_observation_coordinator.py tests/unit/adapters/test_sqlite_observation.py tests/integration/storage \
  tests/integration/service/test_ready_composition.py tests/integration/observation/test_production_composition.py tests/unit/version -q  # 355 passed
uv run pytest tests/unit/adapters/test_codex_rollout_jsonl.py tests/unit/adapters/test_codex_session_stream.py \
  tests/unit/dogfood/test_codex_dogfood_parity.py tests/capability/test_codex_jsonl_import.py tests/unit/protocol/test_import_publication_fixture.py \
  tests/unit/adapters/test_claude_code_integration.py tests/unit/cli/test_claude_code_integration_cli.py \
  tests/unit/cli/test_claude_observe_hooks.py tests/unit/cli/test_cursor_observe_hooks.py -q                                  # 201 passed
uv run pytest tests/unit/application tests/unit/adapters tests/unit/cli tests/unit/kernel tests/unit/dogfood tests/unit/protocol \
  tests/integration/observation tests/integration/storage tests/capability -q                                                 # green after the ripple settled
uv run ruff check <changed files> && uv run ruff format <changed files> && npx --no-install pyright <changed files>          # clean
uv run python scripts/sync_resource_ripple.py --write && uv run python scripts/sync_resource_ripple.py --check              # fixed point
uv run pytest tests/packaging/test_committed_plugin_tree.py -q                                                               # see below
```

Not run here: an installed Codex-testing dogfood (native compaction through the rendered hook, live stream reconciliation under the new admission policy). No real Codex 0.153.4 transcript was available; the #656 fixture matrix is constructed from the 0.150.1 grammar and proves the admission policy, not the actual 0.153.4 event families.

## Boundaries

- [x] No material from `docs/architecture/` or other ignored private drafting inputs is included.
- [x] I will disposition every human and code-review-agent comment before merge: fix it, or reply
      explaining why it is invalid or out of scope.

## Summary

Four fixes so host work on Codex, Claude Code, and Cursor is ingested into the ledger correctly:

- **#657 receipts**: the application classifies what followed an attributable check (responses only, finding-free host observations only, or both) and the receipt says so instead of always claiming finding responses; neutral wording when unclassified; tested boundary stays explicit. No new wire field or gap code.
- **#659 hooks**: mapped-session status probes select their repository locator in a fixed order (explicit project path, host payload `cwd`, hook cwd); the Codex re-ground command renders `--workspace .`; the shared legacy lane no longer probes bare; a companion closed diagnostic names the locator source beside a fence refusal. Missing or conflicting context stays a typed safe failure that keeps the mapping.
- **#619 advice**: observation semantic advice is a durable, generation-fenced background worker (migration 0012). The hook path only looks up or enqueues one row per (workspace, session, evidence basis) with the exact scoped gap tuple; dispatch resolves route, repository authority, and provider binding at attempt time; pending/unavailable are truthful coverage gaps and unattempted review is never success.
- **#656 upgrades**: Codex transcript admission treats `cli_version` as provenance and admits unknown versions under a structural compatibility profile with bounded per-line gaps and a closed admission state; cursors refused under the old policy replay once (mapping `codex-obs-stream/1.4.0`). Claude plugin admission uses the 2.1.233 floor with `tested`/`untested` provenance; Claude/Cursor hook mapping keeps `untested` for unknown versions. ADR-005 and ADR-010 amended.

Model and harness: Claude Fable 5.1 in Claude Code, with two forked sub-agents drafting #619 and #656 (both stopped by the maintainer mid-task; their diffs were reviewed and completed by the coordinator). Work recorded in a Yoetz ledger; receipt summary in the first comment.

https://claude.ai/code/session_01VL3RALUsbJrFmCPSTqVwTC


TheGaySupreme123:
## Yoetz receipt (rcp_afeccfe5-1a3e-4ede-8b6a-0866885e8925, frontier 1517)

**Conclusion:** `unresolved_findings_remain`. All four declared obligations (#657, #659, #619, #656) are resolved; the completion claim records two limitations (no installed Codex-testing dogfood / live compaction; no real Codex 0.153.4 transcript).

**Findings:** 12 hook-observed findings remain unresolved: 6 `failed_work_omitted` (shell command failures during the session, each followed by a corrected successful retry — rejected with that reason) and 6 `stale_evidence_for_changed_state` (verification re-run after the last source edit and published as digest-bound evidence — acknowledged). One recheck did not re-fire them; per guidance no further rechecks were spent.

**Semantic review did not run.** Both checks (whole case and scoped to the claim) returned `semantic_status=failed` / `coordinator_failure` with `semantic_case_content_over_item_limit` on the installed 0.1.0 instance built from `274a74c4`. Coverage is deterministic only.

**Coverage limits:** `content_redacted`, `content_unselected`, `evidence_content_digest_only`, `observation_qualified_partial`, `semantic_case_content_over_item_limit`, `semantic_relevance_review_not_run`.

Verification evidence is digest-only (test stdout and source-diff digests); the full runs are reproducible from this branch with the commands in the PR body.

TheGaySupreme123:
@coderabbit full review

